PR #169, opened 2018-05-27
Attempting to format properly, DO NOT MERGE YET

chuggafan:
This PR is essentially here to do two things:

Format using vscode's default style (yay).

And figure out what went wrong when I was shortening some of the stuff myself, as I consistently am getting errors where I cannot compile ("Out of memory" for some reason), I'd like help to debug as well as to show what areas I think shouldn't be clang-formatted as well as to show what I think are some good changes in format.

I'm mostly making this PR to discuss exactly what the format should be, this is the defualt VSCode CPPTOOLS format from CLANG-FORMAT (aka attempting to be as close as possible to Visual Studio's default)


chuggafan:
For further clang-format options, http://releases.llvm.org/6.0.0/tools/clang/docs/ClangFormatStyleOptions.html should help, but this was done using VSCode's CLANG-FORMAT options, which is very much different from the actual Visual Studio format.

For a further desc. of what I attempted to do: I was trying to compress while loops into macro'd iterated for loops, and taking the beginning of a symbol table, as they took up a huge amount of space and were commonly used.

LADSoft:
ok i'll try to get to this later tonite, trying to fix the bugs in OCIDE right now...

LADSoft:
i like it, your macros are a good idea :)   and c.h looks much better...   haven't looked closely at the source beyond that yet.

 As far as what went wrong when you were shortening stuff, if you take a look for example at line 481 in DECLCPP.c the looping construct was changed probably from a 'while' to a 'for'...   the 'for' now iterates to the next element in the list but meanwhile on line 504 there is a leftover from the 'while' which also iterates to the next element in the list...   in my case it crashed pretty quickly when I fed it a C++ program...   that doesn't seem to be a problem in general but you might want to just double-check every place you added an ITERATOR macro reference...   otherwise my next suggestion is going to be to put parenthesis around used macro arguments (in the #defines)....   probably isn't a problem the way you intend to use these things but it never hurts to be safe.

chuggafan:
Thanks, I had suspected something like that, and the macros were an idea I stole from someone else's repository, it's much cleaner than having to write the exact same thing over and over and is more descriptive. I'll re-look into my macro placements over the next week and hopefully a good standardization of the formatting could be made over that timeperiod

LADSoft:
hi,

well i like what it did with the spacing a lot.   Makes things looks much nicer.   Spaces for tabs is good too.  

Couple of things hit me though as I was reading through source code.

First I never really liked what Microsoft does with 'case' statements.   I always prefered them indented one level...

Second I realized it likes to coalesce long sequences into a single line :).   Mostly that is fine, I don't mind it for function invocations and so forth.   It is just that some of the 'if' statements get incredibly long and sometimes it is more convenient to be able to see bigger chunks of them at once.   

But I don't know if it is viable to have VS format them in a reasonable way.

FWIW some of the wierd formatting in the source code was from an attempt years ago to use an open source code formatter and that effort was sub-optimal at best...

Other than that it looks really good...

chuggafan:
Yhea, indenting case statements, etc. is a definitely possibility, the way I did it was https://releases.llvm.org/6.0.0/tools/clang/docs/ClangFormat.html clang-format the VS style used in VSCode cpptools, if you read the docs [specifically these](https://releases.llvm.org/6.0.0/tools/clang/docs/ClangFormatStyleOptions.html) you can get good formatting solutions from it, I think I'll drop this PR to allow you to make your own then apply my macros changes on top of that (it's a pain in the ass to do it because of how bad I am at regex tbh). Because of the number of changes made to code to make this PR incompatible I also would think that would be better for me personally so I drop as little as possible, clang-format also has the advantage fo working on c AND c++, so that's a good thing.

chuggafan:
I'm currently working on this with a better version with an actual .clang-format file, when I commit it (I should after I do some stuff relating to re-fixing my formatting because I'm terrible at rebasing conflicts), I'll ask again about the different values, I think I know which values I modified which would probably be changed though, but I'll wait until I have an example to actually show it.

GitMensch:
@chuggafan Please don't waste time on rebasing any conflicts here.

As @LADSoft already said he likes the new macros, therefore it would be good to do these as a separate PR first.

Then find a format that you want to suggest, do auto-reformatting only in another branch to show it, may be redone multiple times.
As soon as a good format definition is found: add only the format file and doc/procedures how to apply it (I have the strong opinion that it should be added to the makefiles as recursive "indent" target) as next PR (without the format file actually being applied).

As last step @LADSoft runs the formatter (ideally by only doing `omake indent` on the source main directory) and commits the result.

This way only minimal entries are added to the source history and no/minimal rebasing is necessary.

Note to the formatter: I personally like code being within the 80/100 column range, if separated only by "column overflow" indented more, otherwise by 2-8 spaces (personally actually by tabs with only the "minimal" indentations as spaces but as I've understood @LADSoft this isn't likely to be the style in this repo.

chuggafan:
Yhea, I'll close this pull and redo my macros (that's the longest time taker atm, and since my macros broke things I will be much more careful in reapplication), it'll take awhile though.

Issue #148, opened 2018-05-21
Add Appveyor builds with clang32 and clang64


GitMensch:
Both seem to be supported by both the OrnageC code base and Appveyor.
To verify that the code actually can be compiled and run this way it would be good to include them in the CI-build.

LADSoft:
according to the appveyor build environments page, only CLANG-64 is supported.   Since #170 already added support for it I'm closing this issue.

GitMensch:
Thank you for clarifying.

PR #170, opened 2018-05-27
Adding clang as a compiler to build from on APPVEYOR

chuggafan:
There's no reason this PR shouldn't work except that if clang has more problems than GCC on windows (which it probably does, but we'll find out about that, now won't we). I essentially just copy/pasted from the MINGW64 makefile to edit it up so that it hopefully works properly. Once we can get the warnings all in from clang, gcc, and MSVC that should be a lot of items we could fix and hopefully get less confusing things to happen with.

(this PR is about #148 )

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,7 @@ environment:
   matrix:
     - BUILD_PROFILE: MINGW64
     - BUILD_PROFILE: MS
+    - BUILD_PROFILE: CLANG
  
 build_script:
 # the appveyor virtual machine reported two processors, so, lets try running parallel compile!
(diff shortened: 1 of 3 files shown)


LADSoft:
i already had a makefile for clang???   But it was problematic, the latest version of CLANG was throwing errors it shouldn't have thrown when I tried it again the other day.   I wasn't sure what to do about that...

FWIW at one point I went through and cleaned up a lot of the more distressing warnings on clang, mingw, bcc32, ms... (also could use owc at the time but not since I started seriously using C++ 11 features).   At the time there were also a lot of code incompatibilities that kept the package from compiling on various compilers...

chuggafan:
The way I got CLANG to work was to compile it all with clang-cl (the wrapper for MSVC from clang) and make it 32 bit, then I did a small change to DBGMAIN to make it accept clang as compiling for 32 bit. what went on is clang became more compatible with MSVC moreso than anything else

PR #190, opened 2018-05-31
Adds a macro called __COUNTER__ that increments and removes extraneou…

chuggafan:
…s whitespace from docs

I also added the docs for what the __COUNTER__ macro would be, this extension is actually supported by both [MSVC](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros) and [GCC](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) (and by extension [Clang](https://releases.llvm.org/6.0.0/tools/clang/docs/LanguageExtensions.html#builtin-macros))

I literally added it because I just used it and it's useful in certain situations and can only be implemented by the compiler, it's a neat macro.


chuggafan:
I have no idea why these tests failed as it should do nothing about actual program logic unless it uses the __COUNTER__ macro itsself...

LADSoft:
Thank you for all the documentation updates!

As far as the __COUNTER__ macro looks good, I just wasn't sure what is expected to happen if multiple source files were put on the compiler command line?   As it is the counter won't reset for each file, it will keep counting from where it left off in the last file.

chuggafan:
>As it is the counter won't reset for each file, it will keep counting from where it left off in the last file.

Ah, I had used the methodology of "It isn't resetting between includes", what method would I have to make it reset between source files? Any idea? because the functionality relating to include files is the same across all of the compilers I listed as well

LADSoft:
yeah I would have assumed it would keep counting, through all the includes in a given source.    I just wasn't sure of the behavior when starting a new input file...

To reset it when a new input file starts, there are 'init' functions in many of the modules that get called once for each of the primary files. (but not when an include file is started).   For example the one in the preproc.c module is 'preprocini'.   Many sensitive compiler variables get reset here for each input file...   I would expect if you need to reset the counter for each source file you woudl do it here.

LADSoft:
are you going to change this or do you want me to accept it as-is?

chuggafan:
Well, it seems perfectly fine as is and I did do the changes you asked for, the build was broken due to the fact that I rebased ontop of the broken build, but other than that this PR has really nothing more to add

PR #232, opened 2018-06-17
Fixing some errors found by cppcheck, also commenting on a few more

chuggafan:
I, along with the errors that cppcheck found, in some locations where it was giving a warning I put a comment, where I am questioning or wondering about a design choice, such as:

https://github.com/LADSoft/OrangeC/compare/master...chuggafan:upMaster#diff-16c49de39bc2f0adb93b1ea00fda77c0R1345

https://github.com/LADSoft/OrangeC/compare/master...chuggafan:upMaster#diff-775cc0954e94eec1e16835599e75dff4R1600

https://github.com/LADSoft/OrangeC/compare/master...chuggafan:upMaster#diff-b5fa875d3a8b2f0ae8bdc096796caa96R1063

https://github.com/LADSoft/OrangeC/compare/master...chuggafan:upMaster#diff-ffe719e9c2d4d7f4f8643c7dbd90a34eR1081

https://github.com/LADSoft/OrangeC/compare/master...chuggafan:upMaster#diff-e46b4622d2f75c9381fdc159fd44868bR1691

and finally:

https://github.com/LADSoft/OrangeC/compare/master...chuggafan:upMaster#diff-b78d2d40c114a139ba962eb2279ae105R5109


LADSoft:
I can't build working programs right now....   if I get it working later tonite I will try to look at this.

LADSoft:
Hi chuggafan, will be checking in the 'export class' code as well as a fix for your latest crash soon...  once the CI is caught up I will evaluate this.

chuggafan:
If you could cancel the currently running CI build that'd be great (it appears that oasm has somehow hanged and isn't returning anything). It's especially weird since I haven't changed oasm semantics at all, nor olink, sigh). But yhea other than that it sounds great! My latest two commits are using Visual Studio and rebuilding and going from there, I'll eventually get to applying the CppCoreGuidelines that Visual Studio checks for (and is unique in being able to check for all of them), but that'll take literally forever (last time I tried to build that I got over 10k warnings, ouch)

LADSoft:
I'm starting to look at this now

chuggafan:
You reminded me I had to merge config.c, oops on my part.

LADSoft:
thanks for fixing that!   

the problem with this PR right now is that the resulting compiler it can't compile cregex.cpp; apparently the compiler gets into an infinite loop.  I saw this happen several times...  To see it happen you can compile the compiler, then go to the clibs\cpp\libextra directory and type:

`occ /I..\..\stdinc;..\stdinc;..\libcxx\include cregex.cpp`

chuggafan:
Hmm, any idea which of my changes might be affecting this? I only see renaming mostly in the occ files modified, I'll take a deeper look in a little about it.

chuggafan:
Yhea, so I go to debug this issue you found that popped up, and lex in visual studio says these two things:

Name | Value
--- | ---
... | ...
line | 3615
realline | 3615

I have no idea how something I did screwed it up that bad, but apparently I did, huh

LADSoft:
take a look at cpplookup.c, line 2442.   looks like 'delete' got hit accidentally...

chuggafan:
Yup, seems you're right, ironically, the *ONLY* file that broke when doing further testing (aka made it take forever) was including vector, huh... I'll change and commit that, thanks!

LADSoft:
thanks for the updates, that was quite a bit more invasive than I was imagining!   I didn't proofread every line of it though since we have the CI builds :)

chuggafan:
@LADSoft i didn't get to *NEARLY* as many as Visual Studio was outputting when CppCoreGuideline checks were enabled, that'd take at least a few months to fully go through. I would still like to have you look into those commented lines I put and see if they're valid questionings and whatnot though, as I didn't originally write the code, I would think it's better if you checked them over yourself ;)

@GitMensch Thanks! I don't think I actually do much other than accidentally find bugs and attempt to debug, but I am definitely using all of this to understand compilers, particularly c/c++ ones (clang is quite verbose and understanding LLVM is even more difficult, and GCC is a steaming mess of a codebase)

LADSoft:
@GitMensch will do as soon as I get the codebase reformatting done (should be today or tomorrow)

@chuggafan well when I finally got to the links you put in your first post here they had expired.   I saw a few of them in passing while browsing your changes and fixed the ocide thing where len == len2 (two places) but haven't pushed it yet.   But I will review this in more detail tomorrow :)

LADSoft:
looping past labels and phi objects and so forth in intermediate code:  never results in null pointer because it always stops at an instruction, and there is always a 'ret'.

the peep.c problem: i commented out the relevant code as I'm not sure what the optimization is supposed to be at this point

ideeval.c problem not sure what the comment is supposed to mean?    the code looks right...

window.c problem len == len2 (two places) fixed

PR #246, opened 2018-10-04
Add some fixes for NASM compilation

chuggafan:
Current OrangeC compiler aborts when trying to compile NASM's zerobuf.c because it was "uninitialized" when it had to initialize the array with 0s, this commit is to fix this.

Still haven't figured out how to get omake to work with NASM's autogenerated makefile that I modified to work with OrangeC's build system.

Will also update on my progress for the rewrite


GitMensch:
I vote for inclusion if all current working asm tests still work. We do have some tests, don't we?

LADSoft:
I'm looking at this tonite, don't know if I'll get it done tonite though.

As far as tests the way I *used* to test oasm is compile the entire suite via assembly.    That caught quite a few hard to find issues.    Hasn't been tested that way in a while, but I haven't been making changes either.   

For NASM I may not be able to do that as I don't know that I've been keeping my NASM backend for orange C up to date.    But perhaps I can test it by compiling CC386?

GitMensch:
@LADSoft wrote

> As far as tests the way I used to test oasm is compile the entire suite via assembly. That caught quite a few hard to find issues. Hasn't been tested that way in a while, but I haven't been making changes either.
>
> For NASM I may not be able to do that as I don't know that I've been keeping my NASM backend for orange C up to date. But perhaps I can test it by compiling CC386?

src/occ/parse/regs.h
```diff
@@ -43,8 +43,7 @@ int regmap[REG_MAX][2] = {
     {0, -1}, {1, -1}, {2, -1}, {3, -1}, {4, -1}, {5, -1}, {6, -1},  {7, -1},
 
 };
-static ARCH_REGDESC regNames[] = {
-    {"EAX", 0, 0, -1, 5, {8, 9, 16, 20, 24}},
+static ARCH_REGDESC regNames[] = {{"EAX", 0, 0, -1, 5, {8, 9, 16, 20, 24}},
```
GitMensch: Please use original format/line break in this file. Thanks.

src/occ/x86/gen.c
```diff
@@ -1,4537 +1,4534 @@
...
-            {
-                LLONG_TYPE i;
-                i = apl->offset->v.i;
-                switch (sz)
-                {
-                    case ISZ_USHORT:
-                        i &= 0xffff;
-                        break;
-                    case -ISZ_USHORT:
-
```
GitMensch: These new empty lines are a neat change ;-)

GitMensch: Is there a "coding/formatting style" for this repository already? I'd personally prefer an indent definition file + a complete automated change on all source with a (different) PR and keep it out from those changes.

BTW: I like the zpath change and can review this part as "fine".

chuggafan: > Interestingly the diff online shows only some new empty lines added while the discussion shows that actually some lines were deleted... [I can't comment on the actual remove of those lines...]

I don't know what git did to this commit, however the main changes here shouldn't have affected anything afaik, I think it may have screwed up what newlines are used, though, or something.

src/occ/parse/regs.h
chuggafan: I ran this through clang-format 7 and that's what seems to have messed EVERYTHING up, I have no idea why, but it seems to do so for some reason.

Sometimes I wonder if the dev tools I use are out to get me.

LADSoft:
I tried to manually merge, however, one of the files that changed seems to have had such drastic changes that git considers the entire file to have changed.   It seemed like there was a possibility the line endings had changed (as noted in the last comment).   At this point I can't do the merge...   can you look at it on your end?  (file is expr.c, I was able to do the changes to template.c properly)

chuggafan:
I'll look into it tomorrow, I have a lot of other work today to do.

chuggafan:
Ok, I've resolved all the issues, I'll re-check my commits to see if I missed anything specific

chuggafan:
I've just looked into using this branch with the latest changes from LADSoft, and it appears that something I did here made compiling later versions of LIBCXX (don't ask me how, I'm looking into it), so it might be wise to put this on hold, however it *is* currently compiling the currently used version just fine, so it might have to do with something I've changed elsewhere. *shrug*

chuggafan:
Fixed the bug I found, should be perfectly good now

LADSoft:
i'll give it a try when I get a chance

GitMensch:
@LADSoft Does the current test include compiling and using anything that contains a little more assembler code? If not: Can you add this please?

As I've understood NASM now compiles fine so using the compiled NASM to compile and use the same sources that are used for testing oasm seems like a reasonable thing to do.

Do we need a new issue (or two) to track this?

chuggafan:
We should probably create another few issues to track this properly, if you'd like to create some PRs regarding it while I work on fixing some compilation issues blocking other things that'd be great.

GitMensch:
> if you'd like to create some PRs regarding it while I work on fixing some compilation issues blocking other things that'd be great

It is unlikely that I'll do any work on LADSoft in the near future. The only thing I'll do is to recheck the building of MPIR (had assembler issues, will recheck after next oasm update) and general tests using GnuCOBOL (so far too much errors are in the testsuite if GnuCOBOL is built with and uses occ during the testsuite, will try to check with next release [298 pending commits in master since last release from May 2018]).

PR #247, opened 2018-10-09
This is for some minor issues I found in alloc

chuggafan:
Includes one big thing I found in <io> that exists in stdinc for some reason, as well as a minor removal of a useless line in mtx.c

I want to take this PR to ask a better question: lea.c I see exists and I wonder if it should be updated at all to use a better malloc algorithm, https://github.com/rampantpixels/rpmalloc seems incredibly good for what is needed and supports Windows, although it is c11 based.

I bring this up because I looked around and found out most other C libraries use an entirely different methodology to malloc, most base theirs around some derivation of dlmalloc (which is what lea.c is based on), FreeBSD uses jemalloc (I don't recommend this because it has a lot of logging built into it, which is something I don't prefer because you can't remove it for performance reasons, although it does make it good for diagnostics), OpenBSD is the same, glibc uses ptmalloc2 with adjustments for glibc, MUSL goes off with their own thing, Microsoft calls their [HeapAlloc](https://docs.microsoft.com/en-us/windows/desktop/api/heapapi/nf-heapapi-heapalloc) function (wow, the wonders of the universe, a function in an OS designed to be used as malloc for once)), however, the issue with any of this is:

A. is it necessary? Probably not, I should still do heavy testing before I even go through any of this with the efficiacy of the current solution.
B. would it actually do good? Don't know until I do heavy testing, however it does seem that rpmalloc takes a lot more of the internal win32 systems into account when actually doing allocation fixes.

### DIFF
--- a/src/clibs/alloc/aalloc.c
+++ b/src/clibs/alloc/aalloc.c
@@ -1,47 +1,47 @@
 /* Software License Agreement
- * 
+ *
  *     Copyright(C) 1994-2018 David Lindauer, (LADSoft)
- * 
+ *
  *     This file is part of the Orange C Compiler package.
- * 
+ *
  *     The Orange C Compiler package is free software: you can redistribute it and/or modify
  *     it under the terms of the GNU General Public License as published by
  *     the Free Software Foundation, either version 3 of the License, or
- *     (at your option) any later version, with the addition of the 
+ *     (at your option) any later version, with the addition of the
  *     Orange C "Target Code" exception.
- * 
+ *
  *     The Orange C Compiler package is distributed in the hope that it will be useful,
  *     but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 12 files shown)


LADSoft:
I'm back now...  was stuck out of town for an extra day sigh.

the LEA allocator isn't used by default...   I had problems with it crashing when I ran the IDE with it so I don't use it.   The allocator I DO use is extremely simple and doesn't merge deallocations the way other allocators do, it simply puts them on a list and then reuses them later if it can.   Seems workable but somewhat inefficient.   I haven't tried to find a better solution lately...

LADSoft:
BTW I am going to leave pull requests sitting until I get through with the current problem I'm fixing...   I need to retest everything in both cases and want to have clarity about what I broke 😄 I'm going to start working on that again as soon as I wade through all your messages.

LADSoft:
rpmalloc uses atomics which as you pointed out don't work right now...   I'm just going to work with the pull as-is and maybe make another issue to eventually evaluate rpmalloc :)

PR #266, opened 2018-11-08
Adding and modifying builtins

chuggafan:
I added the following to the intrinsics and removed them from the RTL shims:

__builtin_ctz

__builtin_clz

and their ```unsigned long``` counterparts.

I modified rotate right and rotate left to be more standards conforming (and work with the MSVC names for their equivalents).


chuggafan:
I'm going to add the unsigned long long counterparts as soon as I can figure out how, however, I'd like this to be here to review how I'm going about implementing this and double check my work.

LADSoft:
ok that looks good.   If you want we can speed it up a tad by using the hash table infrastructure in the compiler to do the name searches (wish we had C++ here lol!).

chuggafan:
Yhea, we probably should do so, I also am going to make an issue later relating to the fact that the generated code due to the way the intrinsics are done here (pass through fastcall then directly generated) is unoptimized, and therefore ignores certain things that are overall good for it (due to the way it's done, I believe), if you know of a way I can get the optimizer to pass through the intrinsics that'd be great, otherwise I'll drop an issue down and see what I can do.

Issue #257, opened 2018-10-27
travis-ci build


LADSoft:
now that github can trigger travis-ci builds again, write a build-script for the travis-ci build.   this is ONLY to build once with an alternate compiler and is not intended to compile the compiler with itself.   Challenge is it has to run on linux and should probably use GCC.   It is assumed something similar to the current scripts could be used as omake is somewhat compatible with gnu make.

GitMensch:
You may want to not only define but also add two internal variables to the Makefile and sub-Makefiles:
`$(COPY)` (`copy` or `cp`) and `$(PATHSEP)` (`\` or `/`) and additional in sub-Makefiles: `$(MKDIR)` (`mkdir` or `mkdir -f`)

`$(PATHSEP)` should fix the current issue.

GitMensch:
`$(MKDIR)` should fix the now current issue - the normal unix `mkdir` doesn't create the missing child directories, but `mkdir -p` will; for portability `$(PATHSEP)` would still be reasonable.
And maybe investigate why `$(NULDEV)` evaluates to `NUL` instead of `/dev/null`

LADSoft:
wish i had read your comments before working on it...  could have used mkdir -p.   that $(PATHSEP) idea is looking good about now too lol!   Maybe I will rework that so it is cleaner, after I get further along.

right now my main problem is with goofy things like random inclusions of windows headers and #include file names not having the right case.

chuggafan:
Yhea, that's a lot of why I wanted things such as c++11 full threading support: it makes supporting linux SO much easier because you're removing windows header dependencies

GitMensch:
... actually I was a bit surprised that you do all this in master branch. That is no issue, I just was expecting you'd do this in a ci-branch, do the complete cleanup and then a single commit to master.
But after all those commits it lost its minor importance completely...

> wish i had read your comments before working on it... could have used mkdir -p. that $(PATHSEP) idea is looking good about now too lol! Maybe I will rework that so it is cleaner, after I get further along.

That was the single reason I've posted them... And yes: doing a cleanup in the Makefiles is surely useful as this would lower the "special" cases to a minimum and make them easier to maintain.

> right now my main problem is with goofy things like random inclusions of windows headers and #include file names not having the right case.

In any case it is good that you fix windows-dependencies where they exist by using the correct casing in `include`s and remove them where they aren't necessary like the change to `system()`.

This definitely is worth the work, even in the parts that aren't ci-related!

[...]

Issue #255, opened 2018-10-27
Azure pipelines


LADSoft:
@GitMensch brought up the idea of azure pipelines in #227.   This idea is nice because apparently builds can be done in parallel even in the 'free' model.   Here is a link to the [Marketplace](https://github.com/marketplace/azure-pipelines/)

GitMensch:
With GH Actions now being used - is there anything that this would be useful for or should it be closed as not-planned?

Issue #227, opened 2018-06-13
Bug in numeric conversions


GitMensch:
@chuggafan pointed to tests for numeric conversions at http://www.tybor.com/.

I've actually run the [test programs](http://www.tybor.com/targzip.gz) and think the results should be fixed and afterwards the test programs tbin2dec.c, tsin.c and tint2flt.c should be added to the CI tests.

Note: running tflt2int.c (273 failures reported with current CI build) also fails when compiled with GCC 5.3.0 and 8.1.0 (linked with `-lm` which provide `fetestexcept()` and co) - 165 failures -. Therefore I won't add it to the tests and guess the number of failures may drop when the others are fixed.

Current results:

~~~
$ occ --nologo tbin2dec.c

$ tbin2dec.exe
Info:  'TAILOR' not defined => NOT using any bug workarounds.
Rounding is add 0.5 and truncate.
MIN_WIDTH=509.
Test ( Pass ) {}

$ occ --nologo tsin.c

$ tsin.exe
Info:  'TAILOR' not defined => NOT using any bug workarounds.
sin(355) = -3.01443533594889100e-05 from math library.
sin(355) = -3.01443533594884500e-05 is correct answer.
About 7.09 bits are wrong out of 53.00 total bits.
That is about 135.00 ULPs (Units Last Place) bits error.
Test Failed due to ULPs error being too large.

$ occ --nologo tflt2int.c

$ tflt2int.exe
Info:  'TAILOR' not defined => NOT using any bug workarounds.
This program tests conversions from floating-point types to integer types.
It tests positive powers of 2.0, e.g., f = 2.0**j
It tests both i = +f; and i = -f;
For numbers that are representable, it checks expected value and no exceptions.
For numbers that are NOT representable, it checks for FE_INVALID exception.
This implementation does not claim conformance to IEC-60559 or IEEE-754
Checks of the floating-point exception flags may be meaningless.
Testing conversion from long double to unsigned long long
-f: j=0, expected FE_INVALID, got: 0; i2 is 0X  ffffffffffffffff
-f: j=1, expected FE_INVALID, got: 0; i2 is 0X  fffffffffffffffe
-f: j=2, expected FE_INVALID, got: 0; i2 is 0X  fffffffffffffffc
-f: j=3, expected FE_INVALID, got: 0; i2 is 0X  fffffffffffffff8
-f: j=4, expected FE_INVALID, got: 0; i2 is 0X  fffffffffffffff0
[...]

PR #267, opened 2018-11-09
This fixes my screwup where I forgot to return the index

chuggafan:
Yhea... I should stop being so jumpy about PRs, I had believed I had done the BSR and BSF intrinsics correctly but I had forgotten that I still needed to return a value in the pointer index passed in... oops.... Sorry about that.

This does bring up a good thing though: there needs to be some easy way to add individual files (rather than projects) for regression testing, the current method is really difficult for doing it. I tried to add some tests when I first made my PR but gave up because it requires, like, 3 different files per file to test.

I also take full blame for this screw-up of mine, I had forgotten about testing THAT SPECIFIC section of the _BitScanForward and _BitScanReverse functions.

### DIFF
--- a/src/occ/x86/beIntrins.c
+++ b/src/occ/x86/beIntrins.c
@@ -50,19 +50,24 @@ void promoteToBoolean(AMODE* addr)
 {
     addr->length = ISZ_UCHAR;
     gen_code(op_setne, addr, NULL);
-    gen_code(op_movzx, makedreg(EAX), addr);
 }
 BOOLEAN handleBSR()
 {
     AMODE* al = makedreg(EAX);
-    gen_code(op_bsr, makedreg(ECX), makedreg(EDX));
+    AMODE* ecx = makedreg(ECX);
+    ecx->mode = am_indisp;
+    gen_code(op_bsr, makedreg(EAX), makedreg(EDX));
+    gen_code(op_mov, ecx, makedreg(EAX));
     promoteToBoolean(al);
     return TRUE;
 }
 BOOLEAN handleBSF()
 {
     AMODE* al = makedreg(EAX);
-    gen_code(op_bsf, makedreg(ECX), makedreg(EDX));
+    AMODE* ecx = makedreg(ECX);
+    ecx->mode = am_indisp;
+    gen_code(op_bsf, makedreg(EAX), makedreg(EDX));
+    gen_code(op_mov, ecx, makedreg(EAX));
     promoteToBoolean(al);
     return TRUE;
 }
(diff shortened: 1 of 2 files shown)


LADSoft:
we could add a test project with a makefile that iterates all the C files in the project directory, building them and running them?   Then we could just throw a file in that directory and be sure the tests in it would run.   That would work for simple tests...   for something that would need several files to do a proper test we could later add some kind of support for subdirectories and more automatic building and running...

chuggafan:
That sounds great. I'll add builtin tests once we have that, that way we can have regression tests that are linked to certain issue numbers and manually add them ourselves, I know of quite a few regressions tests we could therefore add... hmm...

LADSoft:
i added orangec\tests\regression.   For each test you need to add two files, a c or cpp or asm file and a cmpx file.   The cmpx file will be compared against the output of running the created executable (and may be empty).   There are a couple of files that I already added :)

GitMensch:
I still recommend to change the current "manual" regression tests to an autoconf generated testsuite.

You just change the test source ".at" (one single file, or more if you want to group those test sources) and run `make check` to let the testscript be created and run (you can run the testsuite script directly, too).

Possible cons: Needs a shell and diff to run, but those are available in Travis by default and in Appveyor (with its installed MSYS+Cygwin environments) and on a developer machine by WSL additionally. There also exists a java based runner that directly process the tests from the source files.
This approach otherwise needs autoconf to let the testsuite be generated (but autoconf is available in the given environments, too)

pro:
* developer just adds/adjust the test source file
* checks stdout, stderr and return code (all these can be ignored if wanted)
* a test can consist of as much files as you want, both source files and "expected" files [normally a diff to stdout/stderr is enough]
* creates a nice test result log for each single test run and also for the complete testsuite
* single tests may be run with passing a keyword, those are taken from the test title and can additionally be specified
* failing tests can easily be re-run after the tested binaries where updated - or after the test sources where adjusted

I offer to create a PR which changes the regression tests to use autoconf generated testscript, *if* both @chuggafan and @LADSoft are open to use something that won't work in plain win32-only environments directly [but even for this case the java based test runner should work fine].

Opinions?

LADSoft:
my main comment is that as it stands, the programs will not run except on windows.    While they compile on travis, there is no possibility they would work properly if run on travis.   While this could be addressed we keep pushing back #227...    I suppose we've proved the programs would work on mingw though?   And the other thing is we've got issue #255 to move this to azure at some point in the future, would this solution work ok on that platform as well?

GitMensch:
Yes, Travis CI originally had no win32 support , "back then" [Tea CI](https://tea-ci.org) was a similar option that supported win32.
But since October [Travis CI added Win32 environment](https://blog.travis-ci.com/2018-10-11-windows-early-release) so it should be possible to use this one to additionally compile on Win32 *and* run autoconf based tests. See #257 for a request to add this environment with a travis multiple os configuration.

GitMensch:
> While they compile on travis, there is no possibility they would work properly if run on travis. While this could be addressed [...]

Is this a "we should have an issue to support GNU/Linux environments"? If it doesn't exists yet I'm all for creating this one, even if it is low-priority (currently there are only 34 open issues that are not related to ocide so another one doesn't harm).

Issue #75, opened 2018-04-19
FR: auto-value for ORANGEC when not set


GitMensch:
It would be nice to have the binaries check for ORANGEC environment value at start and if it isn't set export it to its own environment with path to current executable + "\..".

LADSoft:
I modified this now.   I did change the ORANGEC environment variable, checkin comment not withstanding

GitMensch:
Doesn't seem to work.

Compiling [conftest.c](https://github.com/LADSoft/OrangeC/files/1978051/conftest.c.txt) with ORANGEC set works:

~~~
D:\dev>occ -! conftest.c  -yy
"D:\dev\orangec\bin\olink.exe" -! /mx /c+ /T:CON32  -yy @$$$OCC.TMP
with $$$OCC.TMP=
  c0xpe.o "conftest.o"  "/oconftest.exe"  climp.l clwin.l Running App: "D:\dev\orangec\bin\dlpe.exe" -!  /mCON  "conftest.exe"
~~~

 if I remove the environment variable ORANGEC the result is:

~~~
D:\dev>occ -! conftest.c  -yy
"D:\dev\orangec\bin\olink.exe" -! /mx /c+ /T:CON32  -yy @$$$OCC.TMP
with $$$OCC.TMP=
  c0xpe.o "conftest.o"  "/oconftest.exe"  climp.l clwin.l Error: Input file 'c0xpe.o' does not exist.
        1 Errors, 0 Warnings
        Errors encountered, not creating output file
~~~

LADSoft:
I aplogize, I rushed through this very quickly right before going to sleep...    and got it so wrong it didn't even build on the server.   I think I've got it working in the latest build though.

GitMensch:
Yes, looks much better now. So likely only "PATH" to set in the future :-)

I've also had to set OCC include flags to orangec\include and orangec\include\win32 in the past when compiling from MSYS, is this intended (if not - should it be fixed now)? I can do some tests on this tomorrow and will close the issue afterwards.

LADSoft:
if ORANGEC is set (and it is set now by default) then when the compiler opens up the file OCC.CFG in the same directory as the compiler it should be able to use it to configure the system include paths.   Some of the other tools do similar things with configuration files.

BTW I don't intend to do much tonite, taking a break lol!

PR #270, opened 2018-11-11
update docs for #75

GitMensch:
Just stumbled over this *slightly* outdated part of the documentation. On the second thought you may want to add something like "if ORANGEC environment variable is set it must be set to..." (if this could lead to conflicts).
Actually I'm not sure that there's a reason to keep `ORANGEC` environment variable.., It is still in the sources, isn't it?

### DIFF
--- a/doc/Tools.md
+++ b/doc/Tools.md
@@ -1,11 +1,9 @@
 # Tools - Table Of Contents
-
  
  This document serves as a table of contents, for the documentation on the Orange toolchain.
 
 
 ## General Tools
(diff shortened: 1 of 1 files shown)


LADSoft:
I added your comment about OrangeC environment variable.   It might be necessary to use it sometimes especially when building OrangeC to avoid ambiguity with multiple installs on the path.   I also added skip-commit to the appveyor.yml, so it won't consider .MD and other text files when deciding whether to build...

PR #276, opened 2018-11-11
Add Regression tests for MSVC builtins

chuggafan:
Changed the BZIP regression tests on my side at the very least to use LF instead of CRLF so that it passes properly.

Add function in peep so that you can create registers with a specific size easily, just for ease of use.


GitMensch:
How far is this PR? It looks like @chuggafan did some changes, added tests and everything works - does it? Is there any reason to *not* merge this now?

chuggafan:
PR is fine, only thing that I plan on eventually doing is add *EVEN MORE BUILTINS* and add a header to pull the builtins in from.

Issue #243, opened 2018-07-02
Recursively evaluated constexpr functions do not evaluate to constexpr


chuggafan:
Seems like an oddball one but this popped up into the pipeline when compiling libcxx:

```cpp

using size_t = decltype(sizeof(char));
static constexpr size_t __not_found = -1;
static constexpr size_t __ambiguous = __not_found - 1;
inline constexpr size_t __find_idx_return(size_t __curr_i, size_t __res, bool __matches)
{
    return !__matches ? __res : (__res == __not_found ? __curr_i : __ambiguous);
}
template <size_t _Nx>
inline constexpr size_t __find_idx(size_t __i, const bool (&__matches)[_Nx])
{
    return __i == _Nx ? __not_found : __find_idx_return(__i, __find_idx(__i + 1, __matches), __matches[__i]);
}
```
This shows it off (this is exactly the issue being found in libcxx)

LADSoft:
I'm thinking about rearranging my commitments to get to the assembler possibly this week, but wondered if you needed this fixed before I do that ?

chuggafan:
I'm currently fine with the way it is, it may be a blocking change but I also think I can modify the source to allow for 1. recursively evaluating the stuff and 2. adding a limit on recursive evaluation.

The latter should be easier, but I haven't really been focusing on this task much anyways...

chuggafan:
Ok, so I think I figured out how to fix this and am adding a (hopefully working) regression test for this and the fix in an upcoming PR.

PR #287, opened 2018-11-19
Fixes constant expression variables outside of functions

chuggafan:
This is for #243 and seems to work fine, all I did was set constexpr variables to const, as they are implicitly const, https://en.cppreference.com/w/cpp/language/constexpr


LADSoft:
it looks like the CI failure is some glitch with appveyor, if you agree I will merge this request.

GitMensch:
Whatever caused this, the changes in the sources cannot be taken responsible that the CLANG environment misses a path (this *may* be a timing and parallel build issue though).

chuggafan:
Looks like GitMensch is right in regards to why AppVeyor failed here: the CLANG build environment seems to have failed at cd'ing to the project directory. I don't know if you think this PR solves the issue it purports or not, if you believe it does than I think it's fine to accept it.

LADSoft:
I didn't realize until after I merged this that the error still needed to happen if the constant's expression was not a constant.   I'm going to go ahead and fix it...

PR #292, opened 2018-11-26
All around minor improvements

chuggafan:
Plenty of commits, mostly just getting minor annoyances removed and improved, I did not touch OASM as much as possible as to not interfere with any of the (possible) upcoming assembler changes to deal with the floating point issues.


LADSoft:
thank you for the extensive changes!   I will review soon.   I'm going to wait until you settle out though lol!

chuggafan:
I've done some of my "improvements" and ran into this bug displayed in this commit where nltypes.h cannot seem to call strftime from strftime_l properly *shrug*

chuggafan:
Sooooo, as it turns out it seems, I have run into a compiler bug with time.h changes I've made where the typedef for time_t doesn't register (somehow) and this cascades down, I have no idea how this was triggered, but I know that it's there .-. I'll research on it more in a few days, I have a lot of work to do for the rest of the day & tomorrow

chuggafan:
Ok, As it turns out, I forgot that we currently have neither threading library nor mutex library for c++, oops on my changes to omake there, I'll try to see if I can shim them in temporarily via pulling in bits from a newer libcxx and then have that issue, but it's much better than before if I can pull it off.

chuggafan:
Turns out, that's a no, since I re-ran into a problem in the original place, which is that we would (to be libcxx compatible) have to implement the __atomic* builtins, which we could in theory, but considering the fact that our atomics are currently broken and do not run in testing... well... I'd imagine that's a problem. I'll remove the recursive_mutex stuff I did and hopefully can just add it in later.

LADSoft:
as far as the planned updates to the atomic functionality I've got the assembler assembling all the x64/sse instructions now, just have to verify the byte code sequences for everything.  We'll see how far I get this weekend :)

chuggafan:
I'm going to close this for now and when I feel up to making some commit's and tidying this up I'll go for it.

LADSoft:
it did seem like you were having a lot of trouble with it...

BTW today I duplicated your problem with ORC not working.   Something about not being able to open the headers...   will see about fixing it and making sure the RES files get cleaned up so it gets tested properly...

PR #300, opened 2018-12-11
Minor improvements

chuggafan:
Ok, so I'm doing this PR now for one reason: I need appveyor to build this for me because *SOMETHING* I changed MUST have screwed with *SOMETHING* in oimplib, because locally *ONLY WITH OMAKE* I get errors from the PATH parser thinking that the PATH is a list of filenames sometimes? It's weird and I need to see if other machines have this problem too. It's pretty much impossible for me to debug this.

Also: one of these changes seems to have made it so that either olink generates the .rel incorrectly or dlpe is reading it wrong, but since the command line arguments are so long I can't test *WHICH* in visual studio... ah the wonders of debugging, hopefully this is related to the first thing I mentioned here.


LADSoft:
I see it failed quickly with the rel file problem...    the most common reason for the rel file problem is that one of the external or segment indexes isn't being set, resulting in an invalid fixup (LR record with either X0 or R0 reference).   You should get a diagnostic for that.   If you want to debug it you can set breakpoints in objieeea.c (putlr()) to see what the faulting symbol is and then try to trace it back.    Or if you want I will pull the PR and look at it.

incidentally if you haven't noticed I'm having my own problems with the appveyor builds lol!   I keep making breaking changes...   and now I'm wondering if I've dramatically slowed the compiler with a recent change?

chuggafan:
Well, since the compiler itself has been untouched in this PR, this means that there are three real options: I messed up something in a dependency of the latter two, I messed up something in dlpe, or I messed up something in olink, I've been working on this for three days now and to be honest I've mostly given up on it, I've done 4 different run-throughs, of the possible affecting files and to be honest I have no idea where the issue could be.

LADSoft:
no worries.   I'll grab it in a little bit and see what I can figure out :)

LADSoft:
ok the good news is the code seems *functionally* correct.   At least it works well enough to compile the tools with themselves...   but to get there I had to fix a lot of compiler errors.

the one that threw VS2017 for a loop was your rewrite to util::fatal()...   it still calls vfprintf but the call should have been changed to fprintf because of your rewrite and VS2017 caught this.

there were some problems with the change to cstdio and ctime...   OCC thinks various things now need to be std:: qualified to find them.   (I usually shy away from 'using namespace std;') But I'm not sure OCC isn't broken as VS2017 has no problem with this change.

Another problem I had was with the occ library macro for atoi, it seems incorrect and OCC didn't like code that used it because of that.   I think the proper declaration should be (in stdlib.h):

#define atoi(s)     ((int) __STD_NS_QUALIFIER atol(s))

where __STD_NS_QUALIFIER aliases to std:: in C++ mode and nothing in C mode.

I did finally figure out how to get to the point of the invalid rel file message you were talking about, I had to get the library compiling...  I'll comment on that once I've evaluated further...

LADSoft:
ok so it seems like other than the compile errors your changes to the tools may be ok...   the problem may be in the RTL somehow.   After some experimenting I found I could duplicate the problem with the (working) tools from my branch if I used the RTL from the PR...   I did further experiments with the tools from the PR on the RTL on my branch and it seems ok.

The problem seems to be that something in the RTL is inspiring the linker to generate bad fixups...   this seems like it would be an existing problem that your changes somehow highlighted so I will look deeper tomorrow.

chuggafan:
So I *WASN'T* crazy, thank you! I think once you make those changes I'll pull them into this then modify the stdheaders again so that the .h files aren't dumping into the std namespace, therefore separating them from c++ as much as possible.

LADSoft:
I beat on it for several hours then finally decided the problem is probably with the switch to the c++ style c headers.   Everything seemed clean up to the point I added those changes.  I'll try to figure out more after dinner...

LADSoft:
if you are curious the problem is with the inline functions in 'cstdio'.    I should have it fixed tomorrow nite...   meanwhile it seems to compile the RTL at least if you just comment them out :)

chuggafan:
Ohhh, I see the problem, so you replaced the functioned with macros with their equivalents since the functions themselves had issues some time ago and me changing the way the files pulled out those definitions displayed those problems...

chuggafan:
I *FOUND* the first issue I had talked about, turns out *SOMETHING* I did caused everything after an @ to not be passed in as parameters and rather interpreted in omake... as if it were a variable, I should have at least *THAT* fixed in a recent push soon.

LADSoft:
I pushed my changes.   there may be something wrong with this PR though, I'm not sure if it is that or if my computer is acting up yet...   I had a lot of problems trying to compile the programs in this PR with themselves (with my modifications as well).    Seems like something just ground the disk to death...   and I had to hard reset my computer to continue.   I gave up after about four tries.   but like I say I'm not 100% sure that it isn't something wrong with my computer...

that example for omake doesn't work as-is, gcc won't load cstdio into a C language program.   When I adjusted it though I wasn't able to duplicate your issue...   what compiler are you compiling omake with when you get the issue?

chuggafan:
With the omake example the compiler itsself is a little irrelevant, the more important factor here was the program being made then tested, this appears to be a very, very machine-dependent bug as the changes appeared neither on appveyor nor your machine, and I'm going to look deeper into it soon, but I'll try testing your changes tonight and see if I have any further issues. I'll close this PR for now and re-open it at a later date once I've looked through it some more.

chuggafan:
I found (all) the issues I believe, and I've found out what caused the issues I mentioned previously, it had to do with leftover memory being used and me removing a needed memset...

chuggafan:
*Sigh* my hopes were dashed again, closing this for now until I can figure out why I'm getting multiple errors for some random unrelated things... At least I fixed the problem I had with the .rel files & others...

chuggafan:
Yhea... I ran into an issue where it seems that somehow std::domain error wasn't getting compiled with the code sometimes? I also got some issues where despite debug info being included when doing the makefile, stack traces would be thrown without line/function information, which seems like either a compiler bug or something I did, no clue which :/

chuggafan:
The debug info was supposedly being compiled in (I *TRIED* to do omake clean then omake fullbuild /DWITHDEBUG=YES but that didn't help with stack-traces), but I don't think it worked well because omake clean doesn't actually clean all of the files there, I'd have to test more with a clean copy of the repository, but that's... a bit hard to do. I'm currently using the "puts everywhere" method of debugging, but I'm pretty much done with trying for tonight and will continue tomorrow midday.

chuggafan:
Ok, so, seems like (on my machine) #302 was the one that was breaking it, yay! This should now work. I'll see if it does in the morning, but for now, I'll re-open this for hopefully the final time.

chuggafan:
*Sigh*,  I think I hit myself over the head with my changes to Omake, I'll close this *again* for now, unless you want to see which changes I did that made it not work (It's probably the semaphore changes tbh.)

LADSoft:
if you want to leave it open I'll pull it wednesday and see what I can see...

LADSoft:
the deal with the semaphores is that I used a named semaphore, which was created by the first instance of omake then shared with all instances of omake called by that instance.   This way I could limit the total number of non-omake processes to the specified limit independently of how many instances of omake are running.

It looks like you relaxed this and gave each instance of omake its own semaphore, which means that an unlimited number of sub-processes could be spawned (because the occ makefiles spawn a new instance of omake for each subdirectory in the project).   This is a bad thing and I think it leads to all sorts of intermittent failures.   You just need to supply the name of the original semaphore I think so that it will be shared.

I'm ready to check in my latest compiler but I'm not sure if my changes are going to collide with the work you have done here...   In addition to rewriting the backend I ended up fixing a lot of compiler bugs I found and then I was tinkering with the linker and the assembler as well as the object file library.   If you can get to the point of getting this to work in the next few days I can deal with resolving any collisions.

By the way I'm really glad you asked me to add that multi threading feature to omake; it really is a timesaver :)

chuggafan:
To be honest, if that's the reason, then thanks that you've found it. I did a least (thankfully) think ahead and make 2 constructors and paths for the semaphore, so... that should only be a quick fix.

If you're unwilling to check in a commit because of my PR, don't be, I can always modify the PR to fit the commit and be mergable.

LADSoft:
it wasn't so much I was unwilling, I was just offering to do the grunt work lol!

chuggafan:
It's fine, I'm always willing to do such simple stuff, main thing I have to go through now is testing. Now that we have the new compiler backend though, I do think I'll do some digging on "Improvements" we can make via either compiler intrinsics, improved codegen, or other such things.

LADSoft:
I tried your changes and my latest together on my computer and it was fine... so... I'm going to go for broke and merge this :)

chuggafan:
Hopefully it works, I was planning on waiting for some of the assembler stuff to get out of the way so that I could make a run through on the .c_str(), etc, but I think that can wait for another time.

PR #309, opened 2019-01-03
ocide: more secure check for cwa extension on open

### LINKED ISSUE #258: FR ocide: open cwa as workspace
Doing `ocide my.cwa` opens ocide and inside my.cwa as text file. Expected behavior: ocide is open and has my.cwa as workspace loaded

GitMensch:
fixes #258 

The "problems" with the old implementation:

* "test.cwa.bak" opens as workspace
* "TEST.CWA" doesn't open as workspace

Both should work as expected with the PR (I haven't checked if there are other places which would benefit from something like this...)

### DIFF
--- a/src/ocide/ocide.c
+++ b/src/ocide/ocide.c
@@ -1,6 +1,6 @@
 /* Software License Agreement
  *
- *     Copyright(C) 1994-2018 David Lindauer, (LADSoft)
+ *     Copyright(C) 1994-2019 David Lindauer, (LADSoft)
  *
  *     This file is part of the Orange C Compiler package.
  *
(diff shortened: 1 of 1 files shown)


GitMensch:
> To combat that some of the projects have a new function 'strcasecmp_internal' which duplicates the functionality [...] in MSVC and orange c there is also a function stricmp which does the same thing... all the compilers know about it except gcc/linux, but there is a modification to the build script that deals with that.

So what do you want me to change it to? You're of course free to edit the file in this PR, too :-)

LADSoft:
i went ahead and changed it to stricmp...

GitMensch:
@LADSoft Is this fine to merge now (I think the failing CI builds are unrelated).

PR #313, opened 2019-01-06
improve linux builds

GitMensch:
[VC](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getcwd-wgetcwd?view=vs-2017) defines it as:
```c
// defined in <direct.h>
char *_getcwd(
   char *buffer,
   int maxlen
);
```

[POSIX definition](https://linux.die.net/man/3/getcwd):
```c
// defined in  <unistd.h>
char *getcwd(char *buf, size_t size);
```

Note: In general it would be better to use `HAVE_UNISTD_H` instead of `GCCLINUX`.

### DIFF
--- a/src/occ/parse/ccmain.cpp
+++ b/src/occ/parse/ccmain.cpp
@@ -42,6 +42,12 @@ extern LIST* nonSysIncludeFiles;
 extern "C" {
     char* __stdcall GetModuleFileNameA(int handle, char* buf, int size);
 }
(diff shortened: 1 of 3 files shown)


GitMensch:
@LADSoft after fixing these three files the Travis build runs again so I'd consider it a good thing to merge it :-)

LADSoft:
thanks!

PR #316, opened 2019-01-15
update __builtin_clz

GabrielRavier:
Assuming ___builtin_clz and __bultin_clzl have identical semantics to the ones from gcc and clang this version should be much faster

### DIFF
--- a/src/clibs/complib/386/clz.nas
+++ b/src/clibs/complib/386/clz.nas
@@ -33,18 +33,8 @@
 SECTION code CLASS=CODE USE32
 ___builtin_clz:
 ___builtin_clzl:
-    mov ecx,[esp+4]
-    sub eax,eax
-join:
-    jecxz full
-lp:
-    shl ecx,1
-    jc  done
-    inc eax
-    jmp lp
-full:
-    add eax,32
-done:
+    bsr eax, [esp + 4]
+    xor eax, 01Fh
     ret
 
 ___builtin_clzll:
--- a/src/clibs/complib/386/clz.nas
+++ b/src/clibs/complib/386/clz.nas
@@ -38,10 +38,16 @@ ___builtin_clzl:
     ret
 
 ___builtin_clzll:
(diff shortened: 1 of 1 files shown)


chuggafan:
Oh yes, those two, I replaced them in-code with intrinsics awhile ago and forgot that the assembly for them exists, I remember having had removed them and then leaving the long long versions because I hadn't figured out how to properly implement them as intrinsics.

I do know for a fact that your changes affect nothing though as the assembly ones are not callable due to how the compiler itself works.

GabrielRavier:
Actually this might affect some things - for example take object files built before the intrinsics were inlined - This change could affect them (and removing them might make those have trouble linking)

chuggafan:
True, however considering that I haven't seen such a case yet and that this function used to be only there as an intrinisic for support for LIBCXX... which was changed to being an actual intrinsic, and considering how each release contains major breaking changes, I doubt any recent changes properly would affect anything about that anymore...

GitMensch:
so... If the actual part changed is not needed why not removing it completely?

GabrielRavier:
Deliberately breaking backwards compatibility is probably not the best idea but I dunno

LADSoft:
generally the policy is not to worry about compatibility at the object file/library call level, at least for things specific to OCC.   Obviously standard functions have to remain intact...

As long as things are kept up to date a recompile will fix any problems there...   couldn't get away with that if this was a for sale compiler I suppose but since it isn't...

the PR you made didn't build...   I looked at the build log and it seems you may have inadvertantly cut out some code that was shared between the three different labels in the file.

GabrielRavier:
I updated the long long version (It should probably work). I don't really expect it to get merged, but still if it builds it's better

LADSoft:
thanks for your contribution!

hang on just a little bit...   appveyor is all over the map in terms of time to build these days.   I need to get a baseline before doing any new pushes or merges so I'll just be doing a few builds right now.

PR #317, opened 2019-01-17
Optimized memcpy

GabrielRavier:
Assuming you don't inline all memcpy or don't use a C implementation this should be better than the original (jecxz and loop ugh)

### DIFF
--- a/src/clibs/string/386/memcpy.nas
+++ b/src/clibs/string/386/memcpy.nas
@@ -29,6 +29,7 @@
 [global memcpy_x]
 SECTION code CLASS=CODE USE32
 
+    align 16
 _memcpy:
     push edi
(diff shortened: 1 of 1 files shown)


LADSoft:
hi,

the file does need an 'align' directive in it to tell the linker the the file alignment...   otherwise all those noops are kind of useless because the alignment at the beginning of the file will be wrong.     Beyond that I have to figure out why olib died in conjunction with accepting this PR but can't do it tonite...

LADSoft:
ok i vetted this, there seems to be something wrong with the algorithms but I don't know what.   but one of memcpys in olib seems to be failing for unknown reasons.

LADSoft:
rejected because it broke the build

PR #318, opened 2019-01-17
Optimize images

GabrielRavier:
I use imgbot, a bot that automatically compresses images in a git repo. The compression is lossless, and should thus only contribute to making the repository smaller.


GabrielRavier:
Yeah gonna try to check if any of these images aren't test images and can actually be safely optimized.

LADSoft:
FWIW the tests are in \orangec\tests; some of the tests might work on images.   a lot of the other stuff you touched goes in the IDE help file which I'm just assuming would work :)

LADSoft:
this needs to build before I can accept it...

LADSoft:
rejected because it broke the build

PR #364, opened 2019-04-20
Fix up a lot of minor annoyances I found

chuggafan:
Fix a lot of extraneously defined locals, also did a bunch of std::unique_ptr -> std::shared_ptr to better represent what the pointers are actually doing.

Most of what I've done is: remove locals that are never used, get rid of a lot of signed vs unsigned comparison warnings, and attempt to use std::shared_ptr where it's appropriate (in place of std::unique_ptr, which in the way it's currently used is terrifying and is not at all what it's meant for).

I seem to have created an error of my own or two, but since I did this all at once I sort-of screwed myself on figuring out all the changes.

On another note: I've begun using Visual Studio 2019, and oh my GOD are there a LOT of null pointer deferences, VS2019 is telling me that there's over 811 cases where null pointers can be deferenced. Which would not be good if we encounter them....

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -91,10 +91,6 @@ PenaltyBreakString: 1000
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Left
-RawStringFormats: 
-  - Delimiter:       pb
-    Language:        TextProto
-    BasedOnStyle:    google
 ReflowComments:  true
 SortIncludes:    false
 SortUsingDeclarations: true
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,6 @@ src/.vs
 *.tst
 *.err
 *.out
+.vscode/
+src/*/Debug/
+src/*/Release
--- a/src/clibs/stdinc/float.h
+++ b/src/clibs/stdinc/float.h
@@ -137,6 +137,10 @@
 #define FLT_TRUE_MIN 0x1P-150
 #define LDBL_TRUE_MIN 0x1P-1075L
 
+// This is so that LIBCXX doesn't get to complain at all when you include <limits>
+// We don't support long doubles, but we can at least let it compile anyways
+#define __LDBL_DENORM_MIN__ LDBL_TRUE_MIN
+
 #ifdef __cplusplus
 extern "C"
 {
--- a/src/coff2ieee/CoffFile.cpp
+++ b/src/coff2ieee/CoffFile.cpp
@@ -365,7 +365,7 @@ ObjFile* CoffFile::ConvertToObject(std::string outputName, ObjFactory& factory)
                     relocCount = relocPointer[0].VirtualAddress;
                     relocPointer++;
                 }
-                unsigned offset = 0;
+                int offset = 0;
                 inputFile->clear();
                 while (offset < sections[i].SizeOfRawData)
                 {
--- a/src/coff2ieee/LibFromCoffDictionary.cpp
+++ b/src/coff2ieee/LibFromCoffDictionary.cpp
@@ -55,7 +55,6 @@ void LibDictionary::CreateDictionary(std::map<int, std::unique_ptr<Module>>& Mod
                 files++;
         }
     }
-    int dictpages2, dictpages1;
     int i = 0;
     for (auto& m : Modules)
     {
--- a/src/coff2ieee/LibFromCoffDictionary.h
+++ b/src/coff2ieee/LibFromCoffDictionary.h
@@ -32,7 +32,7 @@
 #include <memory>
 
 class ObjFile;
-class Module;
+struct Module;
 
 struct DictCompare
 {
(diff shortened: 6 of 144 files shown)


chuggafan:
Hmm, so the new error I've popped up seems to be one located in gather_flowgraph in iflow.cpp, where if there are too many labels it seems to be spitting out from the static local array of labels an invalid label location, therefore making it so that it accesses inaccessible memory. Wheee fun debugging times ahead for me.

chuggafan:
Whelp, I have legitimately no clue what I did to cause gather_flowgraphs to fail for profile.c in dos32's RTL as well. The labels seem to be generated incorrectly due to some change I made, it's going to take me awhile to investigate this.... oh boy. At least I know it's 100% caused by SOMETHING in OCC specifically, not anything else...

chuggafan:
Fixed locally the profile.c problem, turns out I accidentally replaced an imin(a, imax(...)) with std::max(a, std::min(...)) oops. I'm going to look further into the sqlite problems, but that will take longer...

LADSoft:
ok sounds like you are doing a lot of useful stuff.   If you need anything let me know...

chuggafan:
Thanks, I'll probably take a week or two (I will have less time for this stuff starting tomorrow) to actually iron out any bug-making-changes I've made, as I'm going through the full git-diff now.... so I have that fun.... yay

chuggafan:
Ok, so as it turns out I think I've found a bug in OrangeC where it is unable to find a match for something in my updated Token.cpp class, throwing out:

```

Error(232)    C:\OrangeC\include\tuple(415):  Could not find a match for 'std::__1::__tuple_leaf<unsigned int=0, allocator<NumericToken>&&&, bool=0>::__tuple_leaf<unsigned int=0, allocator<NumericToken>&&&, bool=0>(allocator<NumericToken>)' in function std::__1::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<NumericToken>&&&>::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<NumericToken>&&&>(__tuple_indices<...unsigned int=0>, __tuple_types<allocator<NumericToken>&&&>, __tuple_indices<..., __tuple_types<..._Tp>, allocator<NumericToken>&>)
Error(335)    C:\OrangeC\include\tuple(415):  Cannot find a default constructor for class 'std::__1::__tuple_leaf<unsigned int=0, allocator<NumericToken>&&&, bool=0>' in function std::__1::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<NumericToken>&&&>::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<NumericToken>&&&>(__tuple_indices<...unsigned int=0>, __tuple_types<allocator<NumericToken>&&&>, __tuple_indices<..., __tuple_types<..._Tp>, allocator<NumericToken>&>)

```

For some ungodly reason, and this is consistent between this branch and the current master branch of OCC. I currently have no clue WHY, but I may be able to investigate this more soon.

LADSoft:
it is hard to see what to do with the various pull requests because all the builds related to the request changes are failing...   are there any of these you need me to look at?   Or do you just want to leave them floating until you get a chance to iron out the build problems?

chuggafan:
Ah, I mostly forgot about this PR, I'd have to investigate what was happening with the builds, I've mainly been working (and hitting roadblocks) on #368 as I found that to be more pertinent.

LADSoft:
looks like something is screwed up in the make files...   when it goes to build lscrtl.dll the linker can't find a couple of files.   The files are just paths without a filename...   there was also some issue with climp.l but I'm not sure that wasn't just fallout from the previous problem.

Do you mind looking at it when you get a chance?   Should be fairly easy to fix...

chuggafan:
I'll look at it sometime in the next week I think, I haven't touched this PR in quite some time so I'm unsure if I can do much.

LADSoft:
if you could at least get the file conflicts resolved I can merge it and try to look at the library issue myself...

chuggafan:
Even that's difficult, declcpp.cpp is easy to resolve conflict for: template.cpp however is not and I have NO IDEA what I've change to make template.cpp so difficult to work with for me. honestly I'd just say "take all from latest head" but that seems screwy at least in git...
Oh well, I can test on my own time.

chuggafan:
Hmm, so I got template.cpp to work fine, but this PR is broken in too many other ways to count.

Honestly I'm rather partial to closing this and starting over again in smaller pieces.

My one main idea with this PR, as a closing, was that std::unqiue_ptr is being used in places with multiple ownership where std::shared_ptr should be used, or actually, regular pointers should be used.

The STL pointers are there to represent ownership, unique_ptrs are there so that only one object at a time can own them. Shared ptrs are where multiple objects can own them and hold ownership all at once. weak_ptrs are there to track ownership and maybe assume ownership when needed (by converting to a shared_ptr). 

Currently everything is a unique_ptr, this should not be encouraged because it requires that we actually have things being owned by what holds it (normally we're storing them in vectors and lists and grabbing ownership from vectors temporarily before throwing them into the world). Inside of lists the links can at least be regular pointers. 

We need to use the smart pointers for what they're designed for more in terms of ownership models instead of using them as convenient destructors. And I guess that's pretty much it in terms of what this PR wanted, it's something that I honestly feel needs to be re-done from the beginning due to how much of a headache I got from just trying to inspect my changes from 3 months ago.

Issue #366, opened 2019-04-29
Unable to match templates with std::make_shared and parameters


chuggafan:
```cpp
#include <memory>
class Val
{
    public:
    Val(int lin) : line(lin) {}
    int line;
};
int main()
{
    std::make_shared<Val>(12);
}
```

one example of the problem that should compile fine, but instead does:

```
Error(232)    C:\OrangeC\include\tuple(415):  Could not find a match for 'std::__1::__tuple_leaf<unsigned int=0, allocator<Val>&&&, bool=0>::__tuple_leaf<unsigned int=0, allocator<Val>&&&, bool=0>(allocator<Val>)' in function std::__1::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<Val>&&&>::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<Val>&&&>(__tuple_indices<...unsigned int=0>, __tuple_types<allocator<Val>&&&>, __tuple_indices<..., __tuple_types<..._Tp>, allocator<Val>&>)
Warning(436)  testnewheader.cpp(10):  In template instantiation started here
Error(335)    C:\OrangeC\include\tuple(415):  Cannot find a default constructor for class 'std::__1::__tuple_leaf<unsigned int=0, allocator<Val>&&&, bool=0>' in function std::__1::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<Val>&&&>::__tuple_impl<__tuple_indices<...unsigned int=0>, allocator<Val>&&&>(__tuple_indices<...unsigned int=0>, __tuple_types<allocator<Val>&&&>, __tuple_indices<..., __tuple_types<..._Tp>, allocator<Val>&>)
Warning(436)  testnewheader.cpp(10):  In template instantiation started here

```

Here's one that shouldn't compile:

```cpp

#include <memory>
class Val
{
    public:
    Val() {}
};
int main()
{
    std::make_shared<Val>(12);
}

```

And yet it produces the same errors as the above snippet of code:

```
[...]

PR #368, opened 2019-05-27
Fix occ crashing with a pair of functions

chuggafan:
OCC crashes at https://github.com/LADSoft/OrangeC/blob/60fcd108fc530b1db25d02af0ee4f1e351d76348/src/occ/parse/declcons.cpp#L1712-L1714 when funcparams->sp is null, this is what is causing the crash with #366 I've fixed the crashing by just having PromoteConstructorArgs exit immediately if funcparams->sp is null, however, in #366 's case it should not be null, and I should investigate more as to why OCC isn't recognizing that the function's overload/constructor exists in this case... I think it's something to do with std::forward, but who knows...

### DIFF
--- a/src/occ/parse/declcons.cpp
+++ b/src/occ/parse/declcons.cpp
@@ -1711,6 +1711,9 @@ static void genConstructorCall(BLOCKDATA* b, SYMBOL* cls, MEMBERINITIALIZERS* mi
                 }
                 if (!callConstructor(&ctype, &exp, funcparams, false, NULL, top, false, false, false, false, false))
                     errorsym(ERR_NO_DEFAULT_CONSTRUCTOR, member);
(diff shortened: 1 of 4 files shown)


chuggafan:
Adding in issue #370 to this PR because both related heavily to crashing and honestly it's much easier to work on this that way...

LADSoft:
I merged it because I wasn't able to duplicate the build problems; I'll leave it open since you are still working on #370 though.

LADSoft:
that was a good thing using std::stack.   I hadn't used it enough for it to be on my radar so I might have done something a little more drastic :)

chuggafan:
Yhea, std::stack and std::vector are my two favorite STL containers because of their simplicity and usefulness.

My only problem with std::stack is that there's no command to both take the top of the stack and pop the stack after getting it.

Issue #377, opened 2019-08-10
Fix FPF usage to not overwrite the VTable pointers


LADSoft:
various parts of the compiler do memcpy/memset/memcmp on FPF structure instances.   These have a VTable so such references aren't kosher.   CLANG flags the issue verbosely.

chuggafan:
Thinking about it, we don't do any FPF class derivations do we? So if we overwrite the VTABLES with anything but a memset to 0 we should be fine because the vtables still point to the same thing (even if the vtable is a different location it's still pointing to the same code). So I think that the prioritization in this case is to not overwrite the vtable so that the calls that it can make don't get ruined. In the future if we ever derive any classes from FPF this will become a much bigger priority to fix but for now I think it's just memset to 0's that are the problem.

LADSoft:
yeah we don't override them.   I'll have to look; not even sure why the vtables are being generated offhand...

chuggafan:
They might be generated because of C++ reasons, but generally it's only virtual functions, but I think we don't generate vtables for it either? Maybe? I'd have to do a thorough check. I think it's just clang being paranoid tbh but you never know.

LADSoft:
fixed in PR #378

LADSoft:
i jumped the gun this wasn't fixed yet...

LADSoft:
it seems my fix for this caused problems with the build so I'm reopening.

LADSoft:
I wasn't able to determine why removing 'virtual' from the destructor would make a difference.   As far as I can tell there is no reason for it should need a virtual destructor. Moving this to the next milestone...

LADSoft:
on a related issue Coverity found some issues that aren't immediately a problem but that should be reworked anyway.   They are marked 'FIXME' in the floating.h/floating.cpp code files

PR #378, opened 2019-08-10
Fix using the mem* operations on FPF

chuggafan:
I notice #377 and checked, as far as I can see there are precisely 2 instances where this happens and I've fixed them fairly simply.  (note: since when FPF is initialized setZero is called I removed the setZero call, if you want me to re-add the setZero call just in case I'm fine with it but IIRC it should still work the same)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,9 @@ src/.vs
 *.tst
 *.err
 *.out
+src/onm/Debug/*
+src/oimplib/Debug/*
+src/obrc/Debug/*
+src/objlib/Debug/*
+src/coff2ieee/Debug/*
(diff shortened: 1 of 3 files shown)


LADSoft:
actually the issue goes much deeper; sometimes EXPRESSION nodes are memset but they contain FPFs as well..   For example on line 2489 of CPPLOOKUP.cpp...   clang flags a whole bunch of these...   if you want to try to fix it that would be MUCH appreciated :smile:

chuggafan:
I'll go into that, so basically I'll find all the instances of memset on EXPRESSIONS and structures containing them as well. if you told me before merging I would have done it in a new commit

PR #359, opened 2019-03-29
Add some more 64 bit intrinsics

chuggafan:
Name is on the tin, I looked into how MSVC implemented _rotr64 and _rotl64 and did just that, if any calling convention problems come up that's something that we need to fix because I did just a dump of the ASM from MSVC (the optimized version, ofc) and then used that, with fastcall calling convention...

PR #380, opened 2019-08-11
Remove enough memset and memcpy calls that clang no longer warns us

chuggafan:
This is a very basic go-through of #377 and is not extremely thorough because there's tons of calls to memcpy and memset and a lot of structures directly or indirectly reference FPFs, but it's all that clang could find.

Summary of what I did: All memset(fpfstructure, 0, sizeof(whatever)) I made to just assigning a new copy of the structure (At function lifetime end it should be freed up in stack terms anyways).

For the memcpy I replaced with an assignment.

For expr.c what happened to cause every line to change was probably line endings got screwed up somewhere and the amount of effort to figure out what screwed up where in regards to line endings is monumental due to VSCode stating that it's CRLF and I am not bothered enough to figure out the cause. But I did modify expr.cpp on line 2222 to remove a memset and replaced it with a call to a QUAD() constructor and assigned it.

### DIFF
--- a/src/occ/middle/iblock.cpp
+++ b/src/occ/middle/iblock.cpp
@@ -282,8 +282,7 @@ int ToQuadConst(IMODE** im)
 {
     if (*im && (*im)->mode == i_immed)
     {
-        QUAD *rv, temp;
-        memset(&temp, 0, sizeof(temp));
+        QUAD *rv, temp= QUAD();
         if (isintconst((*im)->offset))
         {
             temp.dc.opcode = i_icon;
--- a/src/occ/middle/iconst.cpp
+++ b/src/occ/middle/iconst.cpp
@@ -307,7 +307,7 @@ QUAD* ReCast(int size, QUAD* in, QUAD* newMode)
             }
             else
             {
-                memset(&newMode->dc, 0, sizeof(newMode->dc));
+                newMode->dc = _basic_dag();
                 newMode->dc.opcode = i_icon;
                 newMode->dc.v.i = i;
                 newMode->ans = make_immed(ISZ_UINT, i);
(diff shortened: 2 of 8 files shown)


LADSoft:
ok thanks, appreciated.   Whenever it builds I will accept it

chuggafan:
Apparently it had the ability to cause access violations by just assigning it, means that somewhere something by assignment meant dereferencing an invalid pointer, whoops, looks like this IS the better way for this case, sorry clang.

LADSoft:
hm usually assigments are fine unless there is a user-defined one or one is assigning to unreserved memory...

anyway I have several batch files I use to test basically the same things that Appveyor does.   They run compiles three in a row, for example...    Using them is how I got the recent shared_ptr build (#366) to work the first time in appveyor.   It was shall we say a fight to get that far.   

Do you think it would be useful if I name them something sane and check them in?

chuggafan:
If you have some way to batch test all of them the way appveyor does it'll be useful, but it would still take on my machine on the order of 3, 3:30 hours to build it if it's not all running in parallel. It'd be useful for me anyways so that I can test things quickly and easily, and I know that appveyor has their own server thing you can setup now, so either way I might be able to do it. 

Although my current build is failing because I rebased ontop of a broken build, dangit.

LADSoft:
sorry about that, didn't expect simply changing a version number to be this much trouble.   I think I've fixed it so check back in a while...

I uploaded some batch files; they take about 20-25 minutes apiece to run on my home computer.   But that is without running the tests....   I usually run those separately before starting on these as kind of a sanity check that the compiler isn't too broken...   they allow me to get many errors before starting to compile the package with itself.

chuggafan:
Rebased ontop of working builds and force pushed, any problems identified now are solely in the domain of my issue for this PR. I'm working on this only very barely however so...

LADSoft:
ok if you aren't going to work on it can you remove the pull request?   Also remove #359 please

PR #384, opened 2019-08-16
Attribute parsing for C++ syntax

### LINKED ISSUE #379: Support OrangeC attributes in the C++ standard attribute syntax
Basically, in the C++ standard (and in the C2x draft, notably, as seen in the post london 2019 docs http://www.open-std.org/jtc1/sc22/wg14/www/docs/PostLondon2019.htm ( http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2385.pdf for the actual draft) the "standard" attribute way is to do namespace-based attributes, e.g. gcc will have gnu::weak or clang will have clang::noescape, so what we should have is the ability to do the OrangeC specific attributes in a namespace such as occ::zstring for example and that way we can have multiple ways of writing the same thing, but use it as the "standard"

https://en.cppreference.com/w/cpp/language/attributes for reference.

This isn't an immediate issue and I can understand if this if it isn't something easily done due to the internals of the code (also why in the nine hells did WG14 add the ability for \_\_x\_\_ to be the same attribute as x when C++ does not do this? Hopefully a later working draft changes this to be more in like with c++).

chuggafan:
This should fix #379 for everything except C2x support (We should work on that at some point, staying up-to-date on the things should mean that implementing everything as a whole later will be simpler)

I'm actually kinda surprised how I got this done so fast, I didn't expect it to be so easy, it helps when I only had to implement like, 6 different attributes to be in-line with gcc and clang (as far as I can tell, gcc doesn't ACTUALLY list the C++ syntax style attributes it has, but clang lists what they implement from GCC, so I based it off the clang list). Most attributes have no easy way to test them but I know that it should work since I mostly copy pasted code.

Only attribute that we have that we added that doesn't have an implementation now is [[carries_dependency]] and that's because we never implemented it in the first place and we'd need to do some messing with the atomic codegen to properly implement it. (as seen here: https://en.cppreference.com/w/cpp/language/attributes/carries_dependency )


chuggafan:
I think I found the main issue with my PR and why appveyor takes over an hour to build it: because the things are const and declared inside of the function they are initialized and deinitialized every time the function is called, this is obviously sub-optimal, I had assumed that the compiler (all of them tbh) could hoist out the thing and have it declared once and torn down once for the whole program because of the const-ness but that didn't work, oooops

LADSoft:
that is why I went to such effort to declare it 'static' when I first wrote the function :)

chuggafan:
I had actually checked if static did anything and as far as I could tell it didn't sadly, but making it const and manually hoisting it outside of the function DID. So I just did that instead.

chuggafan:
I have no clue why any test is failing because OMAKE sure ain't displaying it to me.

LADSoft:
just saw this - probably it is failing on 'testlinkage'.   Maybe testlinkage isn't returning 0 for whatever reason?    Usually if the response code is not 2 then a program returned something other than 0, or if it is -(large number here) the program probably crashed.   Although it can be a little misleading since we are multithreading the make...

check that quickly - if that isn't it I'm willing to check it in and fix it offline...

chuggafan:
testlinkage.cmpx and testlinkage.tst are being told as "no differences encountered" by fc /b and is returning error code 0. I don't know why it's happening since as far as I can tell all of the file compares and checks are returning successes...

chuggafan:
NVM figured it out, it's because I was doing an error based test and OCC returned 1 because of that that it failed, I need to get a way so that the test results aren't poisoned because of that.... because that's actually the desired outcome for that file (it checks if all the attributes are applied via the same method and checked against extra applications)

LADSoft:
I can't think of an easy way to do that without making it OS specific...

chuggafan:
Yhea.. cuz all of the tests so far have been "no error successful behavior" instead of "error successful behavior". It's something that still needs to be tested but we need to figure out a way to do it. Maybe a wrapper call to test it? IDK

LADSoft:
FWIW I think the way the 'error' tests are written the tests use the make file override that means 'ignore return code' while redirecting the compiler output, then the actual test is done by making sure the compiler generates correct errors/warnings (file compare).

LADSoft:
thanks!

Issue #369, opened 2019-06-03
Integrating the MSIL compiler


GitMensch:
This is a follow-up to #312. If I see it correct the maintained [source code for the msil-compiler is in this repo](https://github.com/LADSoft/OrangeC/tree/master/src/occ/netil) and has an [additional repo](https://github.com/LADSoft/Simple-MSIL-Compiler) with the same content, but outdated and no releases with binaries.

Questions part 1: Is the MSIL compiler compiled in this repo's CI with OCC and verified to at least do "something simple"?
Questions part 2: Can the MSIL compiler's code be compiled with the MSIL compiler yet? Can it be run (as .net) then, too?

Question part 3: Can the MSIL compiler `occil` be integrated into the CI artifacts and the release? If not: Where is the problem?

LADSoft:
#1 I've only recently gotten it to compile with OCC.   The last released compile was with VS2017.   I need to validate the OCC compiled version but there are also new changes I made last year that need to finish validation.    The additional repo has the last working version (compiled with VS2017) so the two repos are currently out of sync.
#2 The MSIL compiler cannot compile itself at this time....   the problem is that it only handles C language code and the compiler/support code is in C++.

#3 I could integrate it easily enough, with the provision it only gets compiled and tested (can't recompile itself).   The limitation is #2, #1 has to be completed before I could do that though.

GitMensch:
Please drop the additional repo, it does not serve a big purpose.
As far as I see it the things to do would be:

* copy over the README.md into a subdirectory of this repo (possibly some other files, too)
* if not too hard: re-tag the two old versions here that match the tags in the other repo (I personally wouldn't mind); if you do please add (the old) binaries to them (maybe only to the newest one)
* make a last commit there adjusting README.md stating where to find the code, mark the repo as read-only [or instead: clean cut with deletion of the repo, the only reason to not do it in my view is the minimal track of the PRs there]

Then run the tests that are available for the MSIL compiler with the compiled version (provides a minimal test bed), ideally with running them.

Close this issue and have inner peace for a cleaner world of code and more integrated tests in this repo.

LADSoft:
hehehe...   well I'll get to it... eventually...   I AM kinda tired of copying files around to release!   gonna concentrate on C++/libcxx8 problems in the near future though...

GitMensch:
[...]

PR #388, opened 2019-08-23
Minor updates concerning OCCIL integration

GitMensch:
I haven't checked if the part of `OCCIL_ROOT` in the README is still true, but I think if it is both the reference and the check should be adjusted to use the common Orange C variables (as outlined in #369 there is no big benefit in having MSIL in a separated installation, so the common settings and paths should be used).


LADSoft:
thank you.   I *think* I caught the OCCIL_ROOT thing and changed it appropriately, will double check and then change the doc.

PR #420, opened 2019-09-20
Remove a whole ton of files no longer needed and maybe directories

chuggafan:
things such as occil_old and a bunch of other files that were added but never in use I've removed because they have no purpose and are unused by any build system we have.


GitMensch:
Looks like a nice cleanup!

LADSoft:
thanks!

PR #524, opened 2020-08-17
Attempts at #include_next for this

### LINKED ISSUE #523: Adding #include_next for the sake of sanity in regards to libcxx
While #include_next is not officially supported by the C standard, it would be useful to implement if only so that libcxx can be truly dropped in so that every time we update a libcxx version (in the future) we don't have to manually go in and delete files that have this. https://gcc.gnu.org/onlinedocs/cpp/Wrapper-Headers.html

This is for my future sanity, and it would get rid of a ton of the startup editing work in relation to the #355 stuff.

I understand why they have it there, but man is it annoying to deal with.

chuggafan:
fixes #523 if it were working, I don't know what's going on exactly but it seems to go into SrchPath once, goes into RetrievePath, runs the do { } while once, then goes through *again* with no descernable reason as to why. @LADSoft this is about as far as I got, I'm not sure how to proceed to get my testing done and I'm not sure how to change the testing harness to work for this anymore *shrug*. The current way I do it is have 2 files of the same name, one checks for a define and the other defines it. Then error for it not existing>


chuggafan:
Oh *@^$ I accidentally did this against the wrong branch, remaking PR

PR #525, opened 2020-08-17
Attempts at #include_next for this

### LINKED ISSUE #523: Adding #include_next for the sake of sanity in regards to libcxx

chuggafan:
fixes #523 if it were working, I don't know what's going on exactly but it seems to go into SrchPath once, goes into RetrievePath, runs the do { } while once, then goes through again with no descernable reason as to why. @LADSoft this is about as far as I got, I'm not sure how to proceed to get my testing done and I'm not sure how to change the testing harness to work for this anymore shrug. The current way I do it is have 2 files of the same name, one checks for a define and the other defines it. Then error for it not existing.

### DIFF
--- a/src/ocpp/PreProcessor.cpp
+++ b/src/ocpp/PreProcessor.cpp
@@ -38,6 +38,7 @@ void PreProcessor::InitHash()
     hash["warning"] = kw::WARNING;
     hash["line"] = kw::LINE;
     hash["include"] = kw::INCLUDE;
+    hash["include_next"] = kw::INCLUDE_NEXT;
     hash["if"] = kw::IF;
     hash["elif"] = kw::ELIF;
     hash["ifdef"] = kw::IFDEF;
--- a/src/ocpp/ppExpr.cpp
+++ b/src/ocpp/ppExpr.cpp
@@ -148,6 +148,35 @@ PPINT ppExpr::primary(std::string& line, bool& isunsigned)
                 }
             }
         }
+        else if (token->GetId() == "__has_include_next")
+        {
+            token = tokenizer->Next();
+            if (!token->IsEnd())
+            {
+                if (token->GetKeyword() == kw::openpa)
+                {
+                    std::string line = tokenizer->GetString();
+                    int n = line.find(")");
+                    if (n == std::string::npos)
+                    {
+                        Errors::Error("Expected ')'");
+                    }
+                    else
+                    {
+                        std::string arg = line.substr(0, n);
+                        define->Process(arg);
+
+                        rv = include->has_include_next(arg);
+                        tokenizer->SetString(line.substr(n + 1));
+                        token = tokenizer->Next();
+                    }
+                }
+                else
+                {
+                    Errors::Error("Expected '('");
+                }
+            }
+        }
         else
         {
             rv = 0;
--- a/src/ocpp/ppFile.h
+++ b/src/ocpp/ppFile.h
@@ -36,36 +36,38 @@ class ppCtx;
 class ppFile : public InputFile
 {
 
-  public:
-    ppFile(bool fullname, bool Trigraph, bool extended, const std::string& Name, ppDefine* define, ppCtx& Ctx, bool isunsignedchar,
-           bool c89, bool asmpp, PipeArbitrator& piper) :
-        InputFile(fullname, Name, piper),
-        trigraphs(Trigraph),
-        extendedComment(extended),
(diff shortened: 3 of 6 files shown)


LADSoft:
i thought include_next was supposed to start one directory in the path list after the directory the current file is located it, but it doesn't look obvious that that was what you implemented.    Can you clarify?

chuggafan:
You are correct, that implementation of #include_next is far more complex than what I implemented here due to me not re-reading it correctly, the way I have implemented it here assumes that there is a maximum of 2 files with the same name that will reference eachother and not if there is 3 or more.

The way I attempted to implement it was going through the directories as normal and skipping the first instance found of the file, since that would be the "original" include, then finding the next file and including that.

chuggafan:
Just checked since GCC and Clang share the same idea of just skipping ahead in the search path, what this means is that I will have to set up a vector or a stack and keep track where in the include path we are for certain items, this is going to be a significantly larger PR than I originally anticipated.

LADSoft:
yeah that is kinda why i didn't want to tackle it immediately myself lol!    Well do you need me still to figure out what is up with SrchPath?

chuggafan:
I've figured out a lot of it, but to properly emulate GCC and CLANG behavior I'm going to have to kill off the FindFIle "first search in the file path of the last included file", at least for include_next's, the way GCC and CLANG do this I *THINK* seems to be by having a list of directories and then just starting out from whatever position on the list the next searched directory is. This setup works by *FAR* the fastest, however is greatly complex given the current setup of the preprocessor, my current idea is to modify ppFile for each included file to simply give the total number of directories traversed to get there, then run through RetrievePath until that number has been reached and then continuing on as normal.

chuggafan:
I've fixed this to work with the way that GCC's cpp does this, but found a different bug on the way for doing this that I don't know how to solve that's unreleated, afaict this is now good-to-go except for tests which I don't exactly know how to operate for ocpp

chuggafan:
@LADSoft Since you're working on the project again, can you take a look at this and verify it works 100% correctly? As far as I can tell and from my testing this PR is complete and I don't want it sitting around for long enough that I forget that.

LADSoft:
ok will try to look at it tomorrow evening :smile:

LADSoft:
I want to report a minor bug.   If you try to do include_next when already at the end of the search path, it crashes...   Other than that it seems functionally correct :smile:

LADSoft:
#526 is also likely a bug in this new code...   I was able to prove that the first directory in the path IS getting searched.   In my case it finds the file referenced in that directory which tries to load from the NEXT directory on the path...   which is where it fails.    Not sure though why this works for OCC but not OCPP...

chuggafan:
#526 I discovered when working on this code, I did not attempt to fix it here because I don't even know where to begin, when I was debugging it for OCPP I discovered that OCPP was not grabbing the first directory starting from the CLI parser, no idea what was up with that.

The minor bug I'll fix along with adding "No source files do not count in include_next", though the latter of the two I don't know how to do.

LADSoft:
I will fix #526 once you get the minor bug fixed and I can merge the PR...

chuggafan:
I'm sorry for not getting back any earlier, but do you happen to have a testcase for the error that you were encountering so that I don't have to go through the reproduction process as in the recent weeks I've been much more pressed for time than normal.

LADSoft:
no problem...   I will get you something tonite :smiley:

LADSoft:
here is a small test that shows the problem.    Basically the issue is trying to use **#include_next** to go past the end of the include path.

[test.zip](https://github.com/LADSoft/OrangeC/files/5308724/test.zip)

LADSoft:
any idea when you'll have time for this?

chuggafan:
Unsure, I've been having a lot of work for many weeks, I did start investigating at some point and will have to deal with Antivirus stuff since I tried updating VSCode again which failed due to AV. So it might take awhile longer, esp. since the next week has a larger than usual amount of work due to outside factors for me, I haven't completely been ignoring this, just that I've been pressed for time on other matters.

PR #551, opened 2020-12-15
Fix include next issue #523

chuggafan:
This fixes an issue with #include_next specifically with the ability to
include files that are in the same directory as the compiled one, this
is an error and is solved by simply checking if the path has been
completed on its go through, if it has been, return that we have an
error (empty filename). I was originally going to do this
overcomplicated error handling mechanism when I realized that:

1. It was way too complex for anything useful

2. It was far too much effort

3. It would not change the root of the problem, that which the original
   compiled directory was somehow being included despite #include_next
   by definition skipping it.

These 8 lines fix points 1, 2, and 3.

I tested this against the q.bat setup given as a dump from the last PR, the error message should probably propagate the information that it failed on an #include_next, but that's also a far more complex thing than this solution is, unless you want me to fix that in this PR, of which I might have to do the modifications I was originally doing.

### DIFF
--- a/src/ocpp/ppInclude.cpp
+++ b/src/ocpp/ppInclude.cpp
@@ -277,6 +277,10 @@ std::string ppInclude::SrchPath(bool system, const std::string& name, const std:
 				// TODO: find a fix that's faster than this, if compiled with any compiler that has vectorization this should be faster than strlen
 				// A possible fix to make this faster is to use std::string but that requires more knowledge than I have atm
 				memset(buf, 0, 260);
+				if (path == nullptr)
+				{
+					return "";
+				}
 				path = RetrievePath(buf, path);
 				if (!buf) // check if we haven't hit the end of all include paths yet, if we have, break and let FindFile handle the rest
 				{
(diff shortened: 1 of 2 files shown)


LADSoft:
i don't think it necessary to say explicitly that it failed include_next, as long as it does fail when it cannot find the file...

chuggafan:
It does at least say it failed to find the file, but I think having it fail saying "Failed to find in include_next" will help with users who don't know what include_next is supposed to do....

LADSoft:
well you might be able to key a more specific error based off which keyword was used to try to do the include...

chuggafan:
Hmmm,  but would that work when the current error scheme is based off in popFile()?

LADSoft:
seems like it is in pushFile()...  and from where pushFile is called you know what the preprocessor keyword was.    So you could pass the token in, or a boolean indicating which keyword is being used.

chuggafan:
You're right, it should be a five second fix then, I'll go do that

LADSoft:
thanks!

PR #552, opened 2020-12-18
Include next fix

chuggafan:
This fixes include_next so that the very end of the buffer of include paths is also accounted for, as I had forgotten that having an empty path does not mean that we have an empty buffer.


chuggafan:
I just realized this is on the master branch, oops, my B

PR #557, opened 2020-12-21
Finally fixes for both with and without include files

chuggafan:
This is a special case to just make sure nothing breaks due to having an empty user search include path during skipIncludeDepth.


chuggafan:
Just leaving a reference to #523 here just to make it clearer what this is about

PR #560, opened 2020-12-22
Fix bug by skipping files properly

chuggafan:
This fixes the bug that reopened #523 this should not be responsible for the:

```
occparse Version 6.0.45.1 Copyright (C) LADSoft 2006-2020
Error( 11)    C:\OrangeC\include\__config(513):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(517):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(521):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(531):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(535):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(536):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(543):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(547):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(551):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(564):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(568):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(572):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(1251):  Syntax error: ) expected
Error( 11)    C:\OrangeC\include\__config(1458):  Syntax error: ) expected
Error( 11)    C:\OrangeC\src\clibs\cmath\cgamma.c(126):  Syntax error: ; expected
Error(437)    C:\OrangeC\src\clibs\cmath\cgamma.c(128):  Invalid use of type 'int'
Error(437)    C:\OrangeC\src\clibs\cmath\cgamma.c(155):  Invalid use of type 'int'
Error( 11)    C:\OrangeC\src\clibs\cmath\cgamma.c(239):  Syntax error: ; expected
Error(437)    C:\OrangeC\src\clibs\cmath\cgamma.c(241):  Invalid use of type 'int'
19 Errors

```

bugs that appeared immediately afterward, I have no idea what these are anyways and I'm currently too tired to look at the code to trawl through a section of the compiler I sadly have become unfamiliar with over the past 6 months.

### DIFF
--- a/src/ocpp/ppInclude.cpp
+++ b/src/ocpp/ppInclude.cpp
@@ -62,7 +62,7 @@ bool ppInclude::CheckInclude(kw token, const std::string& args)
 		std::string name = ParseName(line1, specifiedAsSystem);
 		int dirs_traversed = 0; // this is needed to get #include_next working correctly, the __has_include versions of this don't need to actually keep track tho cuz they don't actually include
 		name = FindFile(specifiedAsSystem, name, false, dirs_traversed);
-		pushFile(name, line1, false, dirs_traversed + 1);
+		pushFile(name, line1, false, dirs_traversed);
 		return true;
 	}
 	else if (token == kw::INCLUDE_NEXT)
@@ -75,7 +75,7 @@ bool ppInclude::CheckInclude(kw token, const std::string& args)
 		std::string name = ParseName(line1, specifiedAsSystem);
 		int dirs_skipped = 0;
 		name = FindFile(false, name, true, dirs_skipped);
-		pushFile(name, line1, true, dirs_skipped + 1);
+		pushFile(name, line1, true, dirs_skipped);
 		return true;
 	}
 	return false;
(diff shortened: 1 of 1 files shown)


chuggafan:
NVM: This apparently broke some stuff, investigating.

chuggafan:
NVM: I figured out the problem, I'm using the wrong pull of the stuff so I'm running into the include path order bug.

chuggafan:
@LADSoft The problem I found when compiling with these changes causing the #include_next fails is that due to the occ.cfg file, the empty %ORANGEC%\include and %ORANGEC%\include\c files are included in some way, shape, or form, which, being empty, causes the float.h to be checked there first. 

I have presumably fixed this, but am investigating more and will most likely finish this PR sometime midday tomorrow.

LADSoft:
FWIW I also tried it with the include files in the 'proper' place and without specifying the library include file paths in the library build, and it still didn't work...    I was willing to believe that maybe since the includes were specified as system includes in the files but I was specifying include paths as user includes things wouldn't work and I should work around it.   But when it still didn't work even with the includes in the 'proper' places I said something...

LADSoft:
@chuggafan as far as the errors you indicated above, I thought I had fixed them in my very latest push.    Did it work a little better when you synced with the latest?

LADSoft:
@chuggafan I tried your latest changes, but they don't seem to have helped....    I tried it both ways....

chuggafan:
Well, there's a reason I said wait until I can get on this today until you pulled ;), I just threw it up to indicate that I'm working on it and what the errors are, I got it to the point of "The wrong include is getting included", and I haven't tested it with a fullbuild yet, so , uh...

chuggafan:
I've gotten *ALL BUT ONE* error down, and this 100% has to do with our usage of the "last include path" scheme, combined with our mix of stuff when compiling our libcxx libraries, it creates a very unique situation where we're using non-system-included include files with system include files. *cries in impossibilities*.

I know technically it's a speed up, but do we still want the last included file's directory to have the include file? @LADSoft this could be a breaking change, so I want to ping you on this and get your opinion before I do anything

chuggafan:
@LADSoft turns out that the current reason that there's *ONE INCLUDE FILE THAT BREAKS* is because of the 

"	// if not there get the file path of the last included file, and search there"

functionality, because what it does is that it goes to the *last included file's assumed directory*, and not their *actual* directory, due to a variety of reasons, but if this functionality were to be changed, I think a lot of the bugs I have been encountering would disappear into the wind, if you can test the next PR without pulling it and modify that functionality for me, I would be grateful.

PR #561, opened 2020-12-22
Fix all problems I know how to fix

chuggafan:
The current issue that I cannot fix is the searching from the path of
the last included file, because with files that go into deeper
directories, this functionality is 100% broken, not that it probably
wouldn't cause problems down the line, due to it skipping possibly
necessary imports and seriously messes with anything related to
\#include_next or any functionality that decides to respect the ordering
of include paths, which this absolutely, totally, completely, destroys.

Removing this functionality is also a breaking change, so if we're going
to do this, I would recommend doing this now, when it rears the problems
in implementing #include_next and anything that relies on system files.

DO NOT PULL THIS PR UNTIL COMPLETELY TESTED!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! <---------------------------------

This PR can work after the bug I mentioned is fixed, the functionality I'm talking about starts on line 246 of the ppInclude.cpp file, and goes to line 263 of the same file. Until this functionality is either fixed or removed entirely, this PR *WILL NOT WORK*, and I don't generally like to make these decisions, because this is a 100% breaking change, I just recommend that to prevent future headaches, removing this functionality in its entirety be considered. GCC does do the "check current directory" scheme for user includes, but we do this too, but this is not the same as the file path of the last included file.

TL;DR: Please consider this breaking change before we go further on this PR.

### DIFF
--- a/src/ocpp/ppInclude.cpp
+++ b/src/ocpp/ppInclude.cpp
@@ -254,7 +254,10 @@ std::string ppInclude::FindFile(bool specifiedAsSystem, const std::string& name,
 		{
 			rv = rv.substr(0, npos);
 			rv = SrchPath(false, name, rv, skipFirst, throwaway);
-			include_files_skipped = current->getDirsTravelled();
+			if (!rv.empty())
+			{
+				include_files_skipped = current->getDirsTravelled();
+			}
 		}
 		else
 		{
@@ -265,7 +268,12 @@ std::string ppInclude::FindFile(bool specifiedAsSystem, const std::string& name,
 	// #include_next does not search this, skip it.
 	if (rv.empty() && !skipFirst)
 	{
-		rv = SrchPath(false, name, ".", skipFirst, include_files_skipped);
+		int throwaway = 0;
+		rv = SrchPath(false, name, ".", skipFirst, throwaway);
+		if (!rv.empty())
+		{
+			include_files_skipped = current->getDirsTravelled();
+		}
 	}
 	// if not there search on user search path
 	// #include_next basically runs through only these two, and maybe the first, if the 2nd here doesn't run, but it's already an inconsistent feature so
@@ -290,9 +298,9 @@ std::string ppInclude::SrchPath(bool system, const std::string& name, const std:
 	do
 	{
 		bool reachedEndOfBuf = false;
-		if (skipUntilDepth && (filesSkipped < totalNumberofSkipsNeeded))
+		if (skipUntilDepth && (filesSkipped <= totalNumberofSkipsNeeded))
 		{
-			while (filesSkipped < totalNumberofSkipsNeeded)
+			while (filesSkipped <= totalNumberofSkipsNeeded)
 			{
 				// Prevent nullptr exceptions and clear out the value of buf
 				// TODO: find a fix that's faster than this, if compiled with any compiler that has vectorization this should be faster than strlen
--- a/src/ocpp/ppInclude.cpp
+++ b/src/ocpp/ppInclude.cpp
@@ -60,6 +60,7 @@ bool ppInclude::CheckInclude(kw token, const std::string& args)
 			define->Process(line1);
 		bool specifiedAsSystem = false;
 		std::string name = ParseName(line1, specifiedAsSystem);
+		const char* breakpoint = name.c_str();
 		int dirs_traversed = 0; // this is needed to get #include_next working correctly, the __has_include versions of this don't need to actually keep track tho cuz they don't actually include
 		name = FindFile(specifiedAsSystem, name, false, dirs_traversed);
 		pushFile(name, line1, false, dirs_traversed);
@@ -253,7 +254,7 @@ std::string ppInclude::FindFile(bool specifiedAsSystem, const std::string& name,
 		{
 			rv = rv.substr(0, npos);
 			rv = SrchPath(false, name, rv, skipFirst, throwaway);
-			//include_files_skipped = current->getDirsTravelled();
+			include_files_skipped = current->getDirsTravelled();
 		}
 		else
 		{
@@ -289,9 +290,9 @@ std::string ppInclude::SrchPath(bool system, const std::string& name, const std:
 	do
 	{
 		bool reachedEndOfBuf = false;
-		if (skipUntilDepth && (filesSkipped < totalNumberofSkipsNeeded + 1))
+		if (skipUntilDepth && (filesSkipped < totalNumberofSkipsNeeded))
 		{
-			while (filesSkipped < totalNumberofSkipsNeeded + 1)
+			while (filesSkipped < totalNumberofSkipsNeeded)
 			{
 				// Prevent nullptr exceptions and clear out the value of buf
 				// TODO: find a fix that's faster than this, if compiled with any compiler that has vectorization this should be faster than strlen


LADSoft:
This code is necessary for #include, but not for #include_next.    At some point I was working with someone who is very familiar with such things, and the is the solution he recommended to deal with a wide number of projects he was trying to compile.

Would it be possible ignore that code when #include_next is being used?    Would that resolve the issue?    

But that seems like just adding more fuel to the fire.    A cleaner solution may be to construct complete search paths out of the system and include paths and replace the procedural nature of FindFile with a single call to SrchPath.    There would be of course separate paths for #include vs #include_next...    the one for #include_next could be created statically before processing files and then you could just rely on it being constant and traverse it in the natural way, whereas the one for #include would be created dynamically when FindFile is called on a #include directive, to deal with things like the path of the current header.

if you think either of those solutions could work let me know and I'll work on it!

GitMensch:
The unknowledgeable third person (me) likes the SrcPath solution, including the "static" one and the "pre-computed" one (which obviously should only precomputed if `#include_next` is seen the first time per file).

LADSoft:
yeah i like that better too.     Sometimes it is better to do a rewrite than throw in another conditional...   and I have a feeling this might simplify things for @Chuggafan too but not sure...

chuggafan:
> Would it be possible ignore that code when #include_next is being used? Would that resolve the issue?

No, the problem persists precisely because it's being used in #include, I have found multiple times that include files are being found in places they shouldn't be according to search path ordering due to this. The "." directory is perfectly fine to find include files from. 

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

The current directory -> user includes -> system includes is precisely the behavior I think that clang also uses because it emulates GCC. 

https://docs.microsoft.com/en-us/cpp/build/reference/i-additional-include-directories?view=msvc-160

MSVC seems weird, they follow the same pattern as us (kind of?) but they don't deal with #include_next like us, and I can near-gaurantee you, that #include_next can, and in fact will, break with this kind of include system.

Currently, I have encountered multiple files that will skip the libcxx include path because the previous file that included them was in the stdinc include path.

I may have fixed this problem anyways, the test program works again, but I'm doing a randomly slow fullbuild that doesn't work quickly for no reason.

I just want to kickstart this discussion for future reference, and if there's a problem we'll be able ti easily and quickly identify it.

chuggafan:
> yeah i like that better too. Sometimes it is better to do a rewrite than throw in another conditional... and I have a feeling this might simplify things for @chuggafan too but not sure...

Sorry, I couldn't get to this due to having to reboot and Windows Antivirus somehow finding ogrep of all things to be a virus threat (I dunno how it determines that). But going off of my latest fullbuild, I may or may not have fixed the problem.... for now, I suspect that this will cause problems in the future and want to at least have the discussion now, as I stated, in case there are problems in the future and that we already have had this discussion and we can formulate and action plan then. I recognize the issue with changing this, which is the entire reason I haven't already done it.

Also, something I'm going to file as a separate bug report is to have a distinct default configuration depending on whether we're compiling c or c++ because we should not be hitting the libcxx include files if we're compiling c IMHO. I think clang and gcc also do this functionality, but this is a separate issue I just brought up and I'll make a separate issue for it.

chuggafan:
This latest push fixes all the problems I could find and an `omake fullbuild /j:8` after an `omake clean /j:8`, as I mention in the commit, this is all that I think I can find via my normal means until we get through the other problems first.

LADSoft:
well if I'm reading this correctly we sort of match GCC.   They search the include directory of the last included file first if it is a user include, but don't search it at all if it is a system include.    We are searching it first for user includes but searching it after the system include path for system includes...    we probably shouldn't search it if we are searching system include path either as it doesn't make much sense to do so.   MSVC seems to do things in a different order, and then they search the paths of all open include files when they are searching user includes...

chuggafan:
We sort of do and sort of don't match GCC, where we do match GCC is that when doing system search paths we first search system search paths, but we also after that check the directory of the previous included file, then we search the current directory, *then* we search the user search path. GCC just doesn't do that.

GCC also does the order differently for user-based includes, for user-based includes it's current directory -> user -> system. We do last included -> current -> user -> system.

while search directories are entirely up to us, I believe that maybe this "nonstandard" relative to other compilers via the last included file system is harming the ability of us to compile projects that are *extremely* specific about includes and include_nexts.

Thankfully, I am also of the belief that any project that actually cares about this is already doing a lot of bad practices that will not work on another compiler, as it requires multiple projects that have the exact same name, include search path, and include file paths to cause this, and if it gets to that point, well.... we can try to accommodate, but they will be stuck with us for that.

Also: oof, this brought up the old q.bat problem... *again*.... will investigate that one now as well.

chuggafan:
For q.bat from the old test for one of the old PRs, is this the desired behavior?

```

C:\OrangeC\test\test>rem works

C:\OrangeC\test\test>occ /Ia;a\b;c q.c
occ Version 6.0.45.1 Copyright (C) LADSoft 2006-2020
Warning a\b\s.h(1): Warning directive: "s.h"
Warning a\r.h(1): Warning directive: "r.h before include_next"
Warning a\b\s.h(1): Warning directive: "s.h"
Warning a\r.h(4): Warning directive: "r.h after include_next"
Warning c\t.h(1): Warning directive: "t.h"
Error: Input file 'c0xpe.o' does not exist.
        1 Errors, 0 Warnings
        Errors encountered, not creating output file

C:\OrangeC\test\test>rem crashes

C:\OrangeC\test\test>occ /Ia q.c
occ Version 6.0.45.1 Copyright (C) LADSoft 2006-2020
Error q.c(1): Could not open using #include_next  "s.h"  for input
Warning a\r.h(1): Warning directive: "r.h before include_next"
Error a\r.h(2): Could not open using #include_next  "s.h"  for input
Warning a\r.h(4): Warning directive: "r.h after include_next"
Error q.c(3): Could not open using #include_next  "t.h"  for input
1 Errors

```

I think this actually is the correct behavior since t.h is not in the include path in this one, the actual bug I found here was that there is only 1 error and not 2 counted, which I *don't* think has to do with me because the error mechanism I use is the same as the main file one.

GitMensch:
> We do last included -> current -> user -> system.

That may be seen "nice" for performance but that easy leads to a brake. Just take GnuCOBOL as example: it ships parser.h and there are different projects that use the same header name and some also install that (which is the reason that with newer MSVC versions and missing GnuCOBOL parser.h you get strange error messages [MSVC's is about XML parsing], while older MSVC compilers just say "include not found"). It is just asking for trouble to search the last include directory first, I highly suggest to drop this feature completely and write that in the NEWS entry for the next compiler version.

chuggafan:
> That may be seen "nice" for performance but that easy leads to a brake.

Agreed, which is why I brought up this discussion. However, I am also of the belief that by having too generic names that aren't separated by at least a directory (e.g. GnuCOBOL should probably modify how includes are done to <gCob\foo.h>) intrinsically will have these problems if you want to support every single include file, as what if someone wants to include MSVC's parser.h while also including GnuCobol's parser.h, #include_next does not guarantee that this happens, because if a *third* program has parser.h, it too, can be found.

I also recognize that we can't actually change how other projects work, but we can at least make a note of bad habits devs will have and if we ever get problems about it in the future we can always just close the issue with a "CANTFIX" message or something similar.

As I have said, this feature already is skipping the C++ include files for libcxx and directly includes the C versions from the C library, whether or not this is good behavior, I have to investigate, but I highly suspect the answer will be a flat out, resounding, "no" when I test it with libcxx. 


Other than this discussion, which should be moved to a new issue instead of hanging out randomly in a PR, this PR is done if there's no issues with the q.bat situation, which upon further analysis, is the correct behavior as far as I can tell.

GitMensch:
We don't have a way in OrangeC to prefix the paths for `""`, do we?
GCC and clang use `-iquote` for that https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html.

chuggafan:
We do, that's /I, our system includes are done via /z since at the time we didn't have more than 1 character switches.

GitMensch:
So most compilers have `-I` for "local" includes and an extra one for system headers - and OrangeC has it the other way around? that sounds like a quite likely break for a bunch of projects during compile.

chuggafan:
We seem to emulate MSVC's features here more than GCC's, though we should make that GCC wrapper/MSVC behavior like Clang does at some point, IMHO it's not relevant to the problem at hand.

LADSoft:
@ chuggafan, that looks like the right behavior.   Yeah the fact that the errors are generated by the preprocessor instead of the compiler itself does make a difference - I should look at that at some point and try to fix it.   I will check this out in a few minutes...

LADSoft:
@GitMensch - I'm not averse to pulling the part where it references the user includes even when it is supposed to be searching the system includes to make this more GCC like, but I do think that changing the part where it looks in the last included directory is risky at best.    And since all the compilers do it in one form or another I would be averse to changing it...

chuggafan:
Let's move this discussion to #563 I've already referenced this there so that we can continue this discussion where it makes sense and with context.

PR #564, opened 2020-12-26
Fixes include_next for empty include path

chuggafan:
It turns out I accidentally forgot a not in my boolean logic,
whoooooooooooops

The logic should only skip the one file if we're not doing #include_next since the empty dir is already skipped if we're #include_next'ing

### DIFF
--- a/src/ocpp/ppInclude.cpp
+++ b/src/ocpp/ppInclude.cpp
@@ -289,7 +289,7 @@ std::string ppInclude::FindFile(bool specifiedAsSystem, const std::string& name,
 std::string ppInclude::SrchPath(bool system, const std::string& name, const std::string& searchPath, bool skipUntilDepth, int& filesSkipped)
 {
 	const char* path = searchPath.c_str();
-	if (path != nullptr && *path == '\0' && !system && !skipUntilDepth)
+	if (path != nullptr && *path == '\0' && !system && skipUntilDepth)
 	{
 		filesSkipped++;
 	}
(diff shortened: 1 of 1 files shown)


chuggafan:
May or may not work, this passed the first section of my tests but has to pass fullbuild before I get anywhere, just throwing this up temporarily while fullbuild runs.

chuggafan:
This doesn't work for some reason apparently, will look into this more.

chuggafan:
Fixes this, as I mention in the last commit, langinfo.h has everything to do with BIONIC, what we need to do is give us the locale for locale.cpp properly, so that's work on our side that we have to do.

LADSoft:
yeah langinfo.h was used heavily in libcxx 8, libcxx10 seems to have moved mostly to xlocinfo.h which is similar but not the same.   I had to redo the locale stuff for libcxx10, not sure at this point how much langinfo.h is being used like i say mostly it has moved on to xlocinfo.h.    hm...   looks like langinfo.h is gone now, that wasn't intentional sigh.

In general locales aren't well supported by this compiler even in the C language though...    there is very basic support but it never really got fleshed out.    xlocinfo.h mostly gives stubs that call non-locale based functions..    Could be cleaned up I suppose but like I say there isn't much at the C level that would support it either.

Thanks will try this commit soon....

chuggafan:
langinfo.h is supposed to be system provided, if we really want we can just copy the behavior of MINGW and WIN32 and make our lives easier... at least for now...

Issue #569, opened 2021-02-18
OMAKE does not search current makefile path for included makefiles


chuggafan:
When running an out-of-tree build:

```makefile
include ./whatever.mak
```

Will not work because omake for some reason sees this as a goal, likely after failing to include the file properly wherever the actual execution is taking place, assuming that the environment starts where $(MAKE) executes.... Which is absolutely not true in out-of-tree builds.

This is what is causing #568 btw, I'm throwing this up as a separate issue as it really is very different and should be focused on separately.

chuggafan:
Figured out why we aren't including: We don't do in-current-directory checking for the existence of the file, unlike in the compiler.... Wonders, may try to figure out how to modify it soonish.

chuggafan:
Upon further inspection: I have absolutely no idea how I'd accomplish this without very large rewrites and structural code changes, the main thing that I think prevents me from theoretically accomplishing this task for now is that there is no way of getting the currently processed file, and even if there was, I have no idea where I'd do the check for current dir processing due to the general code structure.

chuggafan:
Looking at this again, I think it's possible to fix this if we change it so that the preferred path of searching for new dependencies is the same path as the "goal", however I'm not sure if this will break anything else currently related as then we will have to track every goal file's fullpath. Should I attempt to bootstrap this modification on the current system using the current goals or...?

LADSoft:
i think that would be ok...   it currently tries finding it relative to the cwd but usually the goal file is in the cwd if it is going to be including something.  (usually from using the /C switch)

chuggafan:
Yhea the entire reason I made this issue was because this wasn't happening, my original plan was to change the cwd to whatever the file we're working on's directory is but....

I may or may not add an OS wrapper to get the fullpath of a file and utilize that as the "preferred" path. We still have to figure out linux support for omake at some point but that's a different issue entirely.

chuggafan:
Okay so this is actually more difficult than I stated it to be, I actually made an OS::GetFileDirectory function and added Shlwapi.lib for OMAKE only (yay!) and then got slapped in the face by the thing *still not* working, so I'll have to debug why it's not finding pathext2.mak despite my efforts. ompfh.

chuggafan:
[...]

Issue #568, opened 2021-02-18
Getting vscode-makefile-tools to work right with omake


chuggafan:
Microsoft came out with a new way to get intellisense off of makefiles via https://github.com/microsoft/vscode-makefile-tools

After seeing this I of course immediately set out to try it utilizing omake, this didn't work due to a lack of 3 specific arguments: --dry-run, --keep-going, and --print-directory, 

https://www.gnu.org/software/make/manual/make.html#Instead-of-Execution gives a full list of what we need. 

Dry run in particular seems like we just need to create a long command version, --keep-going might be default behavior? But adding that as a thing that we by default have active instead of just having the option to disable seems nice, and --print-directory might need some testing based off of the GNU make, and seems useful to debug the massive near-incoherent mess that is our buildsystem.

chuggafan:
I was able to quickly get this up and running, however I am running into an issue of where the actual command is being run, so I'll just turn this issue into "Getting vscode-makefile-tools to work right with omake" and work on this when I have more time.

chuggafan:
7 long options have been made, however I have run into a problem where if we do not run omake inside of the src directory everything breaks due to the pathext2.mak file being treated as a rule of all things(?). I will investigate this later tonight and see if I can get everything working.

GitMensch:
@chuggafan  I know you are involved with the libcxx - but what do you think about coming back to this one?

chuggafan:
I forgot to mention that this can be closed because I realized that if we just did a directory based one instead (IIRC -C) it worked perfectly, it seems file-based 100% of all make programs fail but directory based we succeed.

GitMensch:
> it seems file-based 100% of all make programs fail

Sounds like a bug in omake in which case there should be an issue. But for "file based make" remember that "omake sourcefile.c" should in most cases just say "no rule to make sourcefile.c".

GitMensch:
The original point of missing parameters `--dry-run`, `--keep-going`, and `--print-directory` along with their short forms seems also still open (according to `omake --help`). So should we reopen this or create follow-up issues?

chuggafan:
> Sounds like a bug in omake in which case there should be an issue.

Nah, I meant omake and gnu make both decide to barf out at the issue here so I'm not questioning it too hard to be honest.

chuggafan:
[...]

PR #570, opened 2021-04-21
Begin adding other make commands to OMake

chuggafan:
This PR is part of #568 and this is only the beginning, not planning on tackling #569 until way later on, this is just the beginning, the main thing preventing me from adding more is that we can't add multiple long-forms for a single short form with our current solution, maybe we can make a superclass that defines that there is an infinite amount of equivalent long-forms for a single short-form?

### DIFF
--- a/src/omake/MakeMain.cpp
+++ b/src/omake/MakeMain.cpp
@@ -42,10 +42,10 @@
 
 CmdSwitchParser MakeMain::switchParser;
 CmdSwitchCombineString MakeMain::specifiedFiles(switchParser, 'f', ' ');
-CmdSwitchBool MakeMain::displayOnly(switchParser, 'n');
+CmdSwitchBool MakeMain::displayOnly(switchParser, 'n', false, "dry-run");
 CmdSwitchBool MakeMain::touch(switchParser, 't');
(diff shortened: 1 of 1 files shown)


LADSoft:
i will look at this after a while.    Meanwhile there are a couple of ways we could handle the long-form problem you posed;    the easiest way is to make the long-form string in the option specification a concatention of long-form strings (using semicolon for example to separate them).     We could also do a more invasive rewrite and make the option specification of the long-form a vector or an array...   the latter would be more readable and might allow easier programmatic additions if that is something we need..

chuggafan:
I'm thinking a modification along the lines of:

```cpp
    template <size_t N>
    CmdSwitchBase(CmdSwitchParser& parser, char SwitchChar, std::array<std::string, N> arr);
```

would work wonders for this, then what we can do is we can do an std::string compare against everything in `CmdSwitchParser`, maybe update it to work on `std::string` so that we don't have to do expensive `strlen` operations and use more stuff out of it and allows us to do basic std::unsigned_map or maybe even a flat_map or some other smaller system used in, say, LLVM, checking instead of our current system for `CmdSwitchParser::Find`

LADSoft:
That sounds good

PR #573, opened 2021-05-26
Fix up adl docs

chuggafan:
This PR does need at a *MINIMUM* one piece of review, line 301 I'm confused because originally it was `~` being unary complement and `~` being unary minus, but that makes no sense, so I put my "assumed" ones there, but `!` being unary minus also doesn't make sense

### DIFF
--- a/doc/adl.md
+++ b/doc/adl.md
@@ -16,7 +16,7 @@ The `Processor` entity gives the name of the architecture.   In the future it ma
 
 The `Coding` entity gives information for assemblers/dissassemblers.   It allows a translation to be performed between a source-code representation  and the coding sequence required by a CPU supporting the architecture.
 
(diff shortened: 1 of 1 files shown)


LADSoft:
will look at this in detail later, looks like at the beginning at least it is mostly cosmetic and typo fixes.    Yeah using ~ for - is kinda hokey but that is actually the way it works...   at the time I was writing it I didn't wish to get lost in parser details so I differentiated unary minus from the normal minus.    Since it isn't too nice I'll go back in a couple of days and see if I can rectify the situation...   that would probably allow me to change ! back into ~ as well...

LADSoft:
thank you for the updates, much appreciated!   Tonite I will probably be able to make time to look into updating the adl compiler to handle unary complement and unary minus in the standard way, then I will accept this...

Issue #576, opened 2021-05-28
is_nothrow_constructible on constructors with no parameters is false when constructor isn't specifically marked noexcept


chuggafan:
When performing the check in the function `nothrowConstructible`, on line 533 of libcxx.cpp, the only check to determine if something is "nothrow" is with `return sp->sb->xcMode == xc_none;`, which will fail when the xcMode is unspecified.

In order to test this I created a file:

```cpp
#define _LIBCPP_CONSTEXPR constexpr
#define _NOEXCEPT noexcept

template <class _Tp, _Tp __v>
struct integral_constant
{
  static _LIBCPP_CONSTEXPR const _Tp      value = __v;
  typedef _Tp               value_type;
  typedef integral_constant type;
  
  _LIBCPP_CONSTEXPR operator value_type() const _NOEXCEPT {return value;}
  constexpr value_type operator ()() const _NOEXCEPT {return value;}
};

template <class _Tp, _Tp __v>
_LIBCPP_CONSTEXPR const _Tp integral_constant<_Tp, __v>::value;

#define _LIBCPP_BOOL_CONSTANT(__b) integral_constant<bool,(__b)>

typedef _LIBCPP_BOOL_CONSTANT(true)  true_type;
typedef _LIBCPP_BOOL_CONSTANT(false) false_type;
template <class _Tp, class... _Args>
struct is_nothrow_constructible
    : public integral_constant<bool, __is_nothrow_constructible(_Tp, _Args...)> {};
template <class _Tp> struct is_nothrow_default_constructible
    : public is_nothrow_constructible<_Tp>
    {};

typedef void* __libcpp_mutex_t;
#define _LIBCPP_MUTEX_INITIALIZER 0
class mutex
{
    __libcpp_mutex_t __m_ = _LIBCPP_MUTEX_INITIALIZER;

public:
    
    mutex() = default;

    mutex(const mutex&) = delete;
    mutex& operator=(const mutex&) = delete;

    ~mutex() = default;

    void lock();
    bool try_lock() _NOEXCEPT;
    void unlock() _NOEXCEPT;

    typedef __libcpp_mutex_t* native_handle_type;
    native_handle_type native_handle() {return &__m_;}
};

static_assert(is_nothrow_default_constructible<mutex>::value,
              "the default constructor for std::mutex must be nothrow");

int main()
{
    return 0;
}
```

This demonstrates the issue, whilst:

```cpp
#define _LIBCPP_CONSTEXPR constexpr
[...]

Issue #372, opened 2019-07-04
Change or modify libcxx so that we have atomics in C++


chuggafan:
Title of the issue is the main focus, we need to work on modifying libcxx or our own atomics system so that we can get atomic features in C++, this work can be done without #355 being completed in terms of compilation, however libcxx does atomics wholly differently from most c++ libs as far as I can tell and decides that clang should have C11 atomics in C++ and then uses the same semantics... don't ask me why it just does. https://github.com/llvm-mirror/clang/blob/master/docs/LanguageExtensions.rst#__c11_atomic-builtins shows what it does, which means that if we'd want to do minimal changes to libcxx just to stay compatible implementing the clang extension to determine if it's lock free would be the best way.

My guess as to why clang/libcxx for atomic_is_lock_free uses the size of the object being pointed to is because of how they implement CPU level atomics for increments decrements etc. 

Getting this done would mean we can remove, say, InterlockedIncrement and InterlockedDecrement from omake or other stuff. It's something simple and needs distinct focus from the rest of the libcxx work, including threading and mutexes. (threading is actually relatively simple in the newest libcxx comparatively, only problem is actually compiling it... which means that's a #355 issue).

LADSoft:
so i looked into this, probably the existing C11 support will be sufficient (although it wasn't immediately clear if I'd have to add new names for  the intrinsicis).   Then add some __has_feature flags to support it...    may look at implementing __has_feature in the compiler so we don't continually have to update the config file dramatically every time we change out the lib...

chuggafan:
Question as I've begun work on this now, for the atomic fences we don't have a separate implementation for thread and signal fences.

According to the C17 standard draft I've dug up, while similar, there is one important difference: The signal fence only will only execute ordering restraints intra-thread between the fence and a requisite signal handler (e.x. anything passed to signal() and raised by raise()).

This is important for proper signal handling for signals such as SIGTERM in multithreaded environments to handle cleaning up resources that don't get freed by the system on process termination (such as named semaphores on linux).

LADSoft:
i don't remember enough about atomic fences to know details, but seems like at a minimum we'd have to do a minor rework to raise() so we know when we are in a signal hander, then add another RTL function to support fencing against that.   Lastly have a compiler intrinsic make a call to that function at the appropriate points....

chuggafan:
[...]

Issue #355, opened 2019-03-01
General libcxx8 issues


chuggafan:
This branch https://github.com/chuggafan/OrangeC/commit/54b1353e3adb2098865401b636c4ad020c8e714d is just a pure libcxx8 replacement of our current libcxx work, with threading enabled, and I fixed the has_feature thing to work better with how libcxx does it and the way the c preprocessor works. 

OCC has an ICE at <utility> line 901, no idea why, and a bunch of issues you'll see that I've made will pop up (all 4 of the recent ones), I removed the "custom" news and am using the libcxx one, it's the same thing, just inbuilt to the library.

other than that, not much to say other than we've only scratched the very tip of this iceberg of issues so far.

LADSoft:
I checked in changes for the recent issues; other than to validate that my smaller example of the problems in #350 works I didn't try to actually use the results of #350 yet.   I figured I will get the chance when I start looking at this lol!

chuggafan:
I've pushed an update that should fix the majority of the "simpler" problems thanks to your work, it actually goes through most of the stuff.

Main problem seems that libcxx likes to treat <charconv> as c++14 even though it's c++17 so that bugs out... Still got some weird other issues going on as well.

Also, it seems like the compiler accepts ```new``` with multiple parameters, however any ```delete``` with multiple parameters is an error, this should be fixed, I do believe that it's simple enough as well, though, I think just removing the extra check (just making sure first arg is a pointer) would do the trick (though that would allow for BS operator deletes, and BS operator news are already allowed though...). That's one bug...

LADSoft:
thanks.  sorry for the delay, been copying my files to a new computer...

A simple compile of <version> results in 96 errors now...   including the ones you indicated above.   Many of them seem related to evaluating using statements too early...   I'll have to look into that.

The compiler is still C++14 BTW, haven't done the C++17 work yet.

I really need to make time for that constexpr issue at some point, the code generation for some supposedly constexpr values is surreal!

BTW I fixed a crash here before I got into the new computer last night: 61226c15c07a48dab6ae3f4d34670bdeae0dd21d.   Apparently some code I put in a long time ago never got triggered until now lol!

LADSoft:
i grabbed at the low-hanging fruit and added __is_abstract to the compiler builtins.   Took out 18 errors - down to 78 now.   I'll look at it closer tomorrow...

LADSoft:
typeinfo.h does not seem to be defining bad_cast?

[...]

Issue #288, opened 2018-11-19
friend operators are only allowed to take 1 parameter


chuggafan:
Friend operators in OCC can only have 1 parameter, this is incorrect according to the standard, https://en.cppreference.com/w/cpp/language/friend shows that you can define a non-member operator inside of a class, and this is causing an issue with upgrading the libcxx version to libcxx7, as the iterators there require this.

LADSoft:
can you clarify what the problem is?   I had previously fixed this because the old version of LIBCXX requires it too...   I just tried this:


    struct aa
    {
	friend int operator<<(aa&, int);
    };
    struct bb
    {
    };
    int operator<<(aa&, int)
    {
    }
    int operator<<(bb&, int)
    {
    }

and it seems to work fine???

chuggafan:
Found it! Super simple to replicate:


```cpp
template<class T>
class thing
{
    template<class A, class B> constexpr
    friend bool operator==(const thing<A>&, const thing<B>&);
}
```
It's the fact that it's ```constexpr``` that's the cause of this rather than the actual friend operator stuff itself, libcxx7's iterators can be constexpr when not debugging which seems to have shown this.

LADSoft:
ok thanks for tracking it down !

PR #578, opened 2021-05-29
Atomic header support 

chuggafan:
This is currently a draft PR for #372 to allow for testing, review, and criticisms of the code I've worked on so far.

Currently I think the next thing I will work on in regard to this will be adding support for the \_\_c11\_atomic stuff from clang, as well as possibly \_\_atomic\_ stuff from GCC, which should be moderately minimal in terms of effort IMHO, just need to add a keyword for all of them and then modify the atomic stuff to simply take it in as applicable.

So far the one thing I don't know about is doing is how to implement signal_fence, but for *now* I can just marshall our current atomic fence machinery and put a warning there that says "We need to actually figure out what we're doing here?"

Edit: 
I forgot to mention that this PR will depend on the completion of the following issue(s):

#576 so that std::mutex's static_assert doesn't kill the compilation
And any others that pop up during the testing of this PR.

Things not in this PR yet: tests that need to be done for the compiler on this, I find that we likely will need extra tests on this PR to ensure nothing breaks again the way it has during this.


chuggafan:
Added the `__c11_atomic_*` stuff, but it's incomplete and stuff still fails to compile, so tests cannot be run.

Once all the compile errors stop I'll start testing.

chuggafan:
Extra note: actually compiling libcxx requires we have GCC's atomic builtins, https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html notably we need __atomic_add_fetch, which has different semantics to __atomic_fetch_add, so we will need to ruckstrap this into the codegen, which is something I'm not particularly enthused about at this point in time and so will probably put off for a later point, but it *IS* a blocking issue for successful compiles so *shrug*

chuggafan:
I also fixed the clang-format for Clang-format 11 and 12 because I was getting annoyed at VS and the like yelling at me

LADSoft:
ok cool :)

LADSoft:
so i will look at this... soon...   got to get my head around the semantic difference you talked about...

LADSoft:
i just reviewed the code, looks good :)

chuggafan:
With the change located in #588 we get *MUCH* further in the compilation but die in an unexpected area. Specifically we die with the stuff defined in __thread_support, which is due to us not compiling the win32 support folder at all (We also need to see if any of the stuff implemented there is useless for us and probably should make a custom orangec support folder so that we can deal with OrangeC's idiosyncrasies).

I *BELIEVE* I've called out that we lacked that before...?

The other thing is for some reason std::__1::__sp_mut::__sp_mut(void*) is duplicated as a public, I will test my quick fix for why the entire __threading_support stuff currently doesn't work though.

chuggafan:
Did some digging: We began copying files when doing mkdir becuase of https://github.com/LADSoft/OrangeC/issues/355#issuecomment-672409972

I was able to via adding 2 new files + a separate command in the LIBCXX ALL (which I should change to a separate dependency that calls the makefile separately and add it in as a separate dep) get it to attempt to compile the thread_win32.cpp however it requires fibersapi.h and a few SRWLock functions of which we currently don't support inside of our headers but have in our .def for kernel32.

I will commit with my current changes now so that if you wish you can check out the changes I've made and see what the issues are currently with thread_win32.cpp

LADSoft:
i fixed the fibersapi support...   now there are a couple of public/extern problems when building LSCRTL.DLL.   Possibly something isn't configured right, but if not I'll look at it tomorrow night

chuggafan:
Ok so with everything the way it currently is we've gotten it down to 2 undefined externals and a doubly defined external:

```
Error: Duplicate public std::__1::__sp_mut::__sp_mut(void*) in module memory.cpp
Error: Undefined External 'std::__1::condition_variable::condition_variable()' in module mutex.cpp
Error: Undefined External 'std::__1::mutex::mutex()' in module debug.cpp
```
No idea how these three are having problems since they should all be linking to each other. Seems to me like there's a problem with our exporting in __config, will investigate this at some point in the future.

chuggafan:
Checking someting: In the modified __config we have we do:

```
#ifndef _LIBCPP_INLINE_VISIBILITY
# define _LIBCPP_INLINE_VISIBILITY
#endif
```
This is seemingly actually causing some of the issues as listed above, is there *ANY* reason why this is necessary for us to have no inline visibility modifiers?

LADSoft:
that was on my radar to fix but honestly I thought I had fixed it.    We definitely want inlined functions when possible for speed reasons lol...   but not having done it I don't know what the repercussions will be...

chuggafan:
I'm completely rewriting how the __config file works to be more inline with LIBCXX's style *anyways* so I think I'll just remove our special casing here.

chuggafan:
SO: changing the visibility attributes entirely has removed 2 undefined externals and added 2 undefined externals elsewhere that didn't have issues before whilst still having a duplicate public.

I'm definitely feeling like I'm playing programming whack-a-mole at this point in time.

LADSoft:
yeah, still seems a little touch...   I will look into it as soon as I figure out what is up with the atomic builtins :)

chuggafan:
I figured out a different issue with the atomic header because we weren't setting a specific define as well, so will fix and update as this goes on.

I feel like I'm learning a lot more about libcxx's internals right now than I really would ever need...

chuggafan:
The issue with atomic was partially that we were being found as a freestanding platform because we didn't define `__STDC_HOSTED__` if we were in c++ mode, fixed that.

chuggafan:
Even better results from careful analysis of our stdatomic.h vs <atomic>, our memory order is the same in terms of what we start and what we end with, except that our memory_order_relaxed starts at one and ends at 6 and the memory order in <atomic> starts at 0 and ends at 5.

🎉 more bugs found whilst researching the current issue

LADSoft:
i'm about to push a change; it changes the intrinsinc `__atomic_modify` to `__atomic_fetch_modify` and introduces a new intrinsic `__atomic_modify_fetch`.    That should be all you need as a template for creating the new intrinsics you want to create.   If you want you can also create explict intrinsics for the things  `__atomic_fetch_modify` refers to and get rid of `__atomic_fetch_modify` and `__atomic_modify_fetch` entirely...

this also fixes the snafu where the memory order constants were off by one...

chuggafan:
Should be pushing the updated thing in about an hour, a quick ask: Can you cancel all the appveyor builds so that the tons and tons of pushes made recently don't clog it up?

LADSoft:
ok i deleted them :)

chuggafan:
Quick question, when looking at the __atomic_compare_exchange_n function definition for the GCC's builtins they have this fourth parameter boolean for STRONG/WEAK, is the method for getting this boolean doing an expression_assign with a null 5th parameter and checking if the fourth parameter is of the boolean type (This I am also not quite sure of), or do I do something else?

The strong/weak volatile semantics don't really matter to us at this point in time since we are on x86 but if we ever port to ARM it begins to matter as their semantics are much... weaker... than the x86 platform.

LADSoft:
call optimized_expression() which will parse the expression then run it through the high-level constant optimizer to get the result you need.   The atp argument can be null since you aren't doing an assign but the other arguments should be filled in.   The **tp argument will return a pointer to the type associated with the expression, if you want you can check it is an integral type by calling isint() but that probably isn't strictly necessary here because of what I'm going to say next.   The **exp argument will return a pointer to an expression which has been fully constant-evaluated, you should call isintconst() on it to make sure you did indeed get a constant.   If that passes then the expression should point to an expression node with one of the en_c_xxx node types and you can just use exp->v.i to get its value, which will be the bool you specified as the argument to the builtin.

hope that wasn't too wordy :smile:

chuggafan:
I guess that works, this is technically all going to be used as a throwaway but I still need to parse the thing to actually get to the point I want.

chuggafan:
Got it all building, etc. and get slapped in the face with a valid expression giving me "Expression syntax error".

I'll push my changes, this came so far out of left field I don't know why this is going on and any debugging I would do would be fruitless.

The issue appears with locale.cpp on line 665 passing into the atomic_support's __libcpp_atomic_add on line 84.

What an issue... what an issue...

chuggafan:
Also looking ahead at this appveyor thing: filesystem apparently is supported by libcxx if you aren't using C++03, should we just modify the system header for filesystem or just deal with `<filesystem>` as well? The latter will be an absolute massive headache. And the reason this is an issue is that `<fstream>` pulls it in... and `<filesystem>`  also feels special in breaking a huge amount of the compiler in ways that I thought were previously fixed (#288 has begun its haunt yet again with an inline friend bool == operator....)

Also: I don't think we have a way of excluding from explicit instantiation some things, the way LIBCXX decides to go about hiding things from this on platforms that don't explicitly support a specific attribute is to simply force inlining, of which I don't think we have a method to do (we "technically" support __always_inline__ but as per the comment there that's just an extra "tick the optional inline please" thing, but I'm not sure of its efficacy in excluding something from being explicitly instantiated)

LADSoft:
usually when I'm debugging an error, i set a breakpoint in the debugger in ccerr.c say on line 875.   Then I look at the call stack to see where the error was thrown then I can figure out why.   Looking at lex->data gives you an idea of what the token was right before whatever threw the error..

we can add <filesystem> in the next milestone if you want, i think it isn't valid until C++17 anyway?   Write an issue for it...  as for  forcing inline, I can make it force in more cases with an additional flag, but there are still issues where we don't support return structures from inlined functions and we don't support raw structures for inlined functions (for example) so it still wouldn't be perfect...

LADSoft:
i can possibly fix the problems that make always_inline not work...   that will work especially well if we also do #591...   if you think it important write it up and we can also do it in milestone #4.

chuggafan:
TBH we don't *NEED* __always_inline__, what I think we need is some way to force that there's no explicit template instantiation

https://releases.llvm.org/10.0.0/tools/clang/docs/AttributeReference.html#exclude-from-explicit-instantiation

This may actually help with our linker problems...

I'll also just #define out the entire `<filesystem>` header and at a later point we'll work on the issues present, maybe punt the question to milestone 4 for now as we already have enough stuff on our plate....

chuggafan:
So: began tackling the problem with the __libcpp_atomic_add thing, and it turns out the issue is deeper than I thought, we're grabbing an "optimized_expression" and tp1 sadly doesn't get set, however we do get a return type, but the expression we made is also... not the best...

line 2225 of stmt.cpp supposed to be an *optimized_expression* or is it supposed to be a regular expression or an expression_assign? I feel like optimized_expression unless it does some trickery to flatten types to get the correct one we want is... I'll checkout why tp1 is not being assigned under optimized_expression later but this is just my first impressions.

LADSoft:
optimized_expression just calls a variant of expression anyway...   for a regular int value like the one in your example,  expression_primary should be creating a simple integer node and returning a pointer to the stdint structrure...   and that would eventually be returned by optimized_expression() so optimizing is moot

but another issue I forgot about that you might need to cover is that if you are in the middle of parsing a template definition then the expression will be sub-optimal and the type may be either hard to use or invalid...

```
if (!templateNestingCount || instantiatingTemplate)
{
    should expect expression to be a valid constant
}
else
{
   fudge some default
}
```

chuggafan:
Ok so another issue that's apparently going to be needed to fix is in my recently added `__ATOMIC_FOO` scheme where our overloads can't detect the difference between std::to_string(int) and std::to_string(unsigned int).

This case will require some investigation on my part since I'm not sure if enums are by default signed or unsigned... if it's undefined then I say we roll dice and pick a logical one.... if there's any negative numbers we for sure will be forced to select signed.... otherwise unsigned works....

chuggafan:
Ok cool the C17 PDF clears this right up:

All enumerations shall be an interger constant expression that has a value re presentable as an `int`, will check in on the C++17 working draft to see what's up with them next.

chuggafan:
Ok, so here's a comment categorizing all of the issues we currently face:

Linker issues from `locale.cpp` and `memory.cpp`, `locale.cpp` with not exporting two things that may or may not actually need exporting and `memory.cpp` for somehow double dipping in its exports and exporting `__sp_mut` twice. #593 

Enums are not automagically treated as type `int` #596 

Neither of these are particularly of my knowledge set at this point in time as:

1. I don't know enough about how the entire linker setup is to do anything.
2. I have no idea just how many places I'd have to locate and pull out a "cast it to int if you're not aware of an underlying enum type"

LADSoft:
i'm working on #593 now...

chuggafan:
On my side EVERYTHING except `__sp_mut` being defined twice has been *FIXED* that I can immediately test with a fullbuild. YMMV on any other issues popped up by this.

I *STILL* have no idea why __sp_mut is being defined twice but the work on #593 did wonders.

LADSoft:
i just got done with #596, will look at this soon...

LADSoft:
i just pushed; as far as I know I'm caught up on all your issues.

chuggafan:
EVERYTHING COMPILES! 🥳 now it's time for the tests to be done, turns out there's a lot of stuff that has issues now with other things:

```
Error: Public 'std::__1::__cloc()' was also declared as an imported function
Error: Public 'std::__1::__libcpp_thread_get_current_id()' was also declared as an imported function
Error: Public 'std::__1::bad_weak_ptr::~bad_weak_ptr()' was also declared as an imported function
Error: Public 'std::__1::cout' was also declared as an imported function
Error: Public 'std::__1::ios_base::~ios_base()' was also declared as an imported function
Error: Public 'std::__1::ios_base::__set_badbit_and_consider_rethrow()' was also declared as an imported function
Error: Public 'std::__1::ios_base::failure::~failure()' was also declared as an imported function
Error: Public 'std::__1::locale::~locale()' was also declared as an imported function
Error: Public 'std::__1::locale::has_facet< const (*)(locale::id&) const >' was also declared as an imported function
Error: Public 'std::__1::locale::use_facet< const (*)(locale::id&) const >' was also declared as an imported function
Error: Public 'std::__1::system_error::~system_error()' was also declared as an imported function
Error: Public 'std::domain_error::~domain_error()' was also declared as an imported function
Error: Public 'std::invalid_argument::~invalid_argument()' was also declared as an imported function
Error: Public 'std::length_error::~length_error()' was also declared as an imported function
Error: Public 'std::logic_error::logic_error(char const *)' was also declared as an imported function
Error: Public 'std::logic_error::logic_error( const logic_error&)' was also declared as an imported function
Error: Public 'std::logic_error::~logic_error()' was also declared as an imported function
Error: Public 'std::out_of_range::~out_of_range()' was also declared as an imported function
Error: Public 'std::overflow_error::~overflow_error()' was also declared as an imported function
Error: Public 'std::range_error::~range_error()' was also declared as an imported function
Error: Public 'std::runtime_error::~runtime_error()' was also declared as an imported function
Error: Public 'std::underflow_error::~underflow_error()' was also declared as an imported function
```

LADSoft:
ill look at this now...

chuggafan:
Uhhh why is this closed?

This is still an active PR....

LADSoft:
i guess i hit the wrong button lol...

did you catch that i added __has_attribute to the __config file?

chuggafan:
Oh! Maybe that's the issue... I'll check

chuggafan:
Retested after adding __has_attribute, still issues... sigh...

chuggafan:
Running tests apparently we're having issues with stdatomic-lockfree tests as well. So there's a lot we need to go out and verify...

LADSoft:
i fixed some problems with compiling the library; now lscrtl.dll links properly.

the definition of __has_attribute is like this:

'''c++
#define __has_attribute(__x) (__has__##__x)
'''

which is a little different from the other __has definitions.   Not sure how it is using the linkage stuff without that being correct though...

also, I tried a quick test program against the lib:

```c++
#include <stdexcept>
#include <stdio.h>
int main()
{
	std::logic_error aa("hi");
}
```

which results in the same sort of import errors...   I generated an assembly language file and looked in it and it seemed like there were import statements for all the library stuff.   hence the error about colliding with non-imported versions.  Not sure if that is a header issue or if the compiler is broken, will look at it some more tomorrow...

there are also missing externals, that might be a compiler bug of some sort...

LADSoft:
yeah getting the definition of _LIBCPP_DLL_VIS correct fixed the remaining link problems at least for normal links.   It has to be blank for a normal link, or declspec(dllimport) if importing from lscrtl.dll...

i'm not totally surpised if some of the atomic builtins I added don't quite work...   I was only able to do very basic testing...

LADSoft:
there is also some issue where if you drag in the locale support (for example using io streams) it crashes during construction of the locale object.    Not sure why yet, just know it doesn't happen if you disable the threading support...

chuggafan:
Pulled in the new stuff and fixed the `__has__` thing, still have the same issue sadly, just with an extra:

`Error: Undefined External 'std::__1::basic_string<char, char_traits<char>, allocator<char>>::assign( const basic_string<char, char_traits<char>, allocator<char>>&)' in module generator.cpp`

*sigh*

Also: the atomic builtins not quite working can be fixed, we can go steal more of gcc and clang's tests related to them and execute them to see if they run.

LADSoft:
hm it seemed to work for me, will try again ...

chuggafan:
https://github.com/LADSoft/OrangeC/commit/14ba9c09d424bdbe18264fffdc9fb81ff441123d

looking through this again... it seems that our way of doing __has_attribute is unlike https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html

It would probably be better to knock out the extra `__` on both sides for this and maybe make it so that it gets handled somewhere in the parser as then we could also add the compiler-specific attributes... because now that I look it's likely we're not having it queried correctly.

LADSoft:
I did a rebuild and everything is working..

You do also need to change the definition of _LIBCPP_DLL_VIS on line 664 of __config.   I just set it blank for testing but It should probably be something like:

```c++
#if defined(__ORANGEC__) && !defined(__LSCRTL_DLL)
#define _LIBCPP_DLL_VIS
#else
#define _LIBCPP_DLL_VIS declspec(dllimport)
#endif
```
assuming you are supporting other compilers and want it always set for those.

because occ generates different code if you are using imports than it does otherwise.   here __LSCRTL_DLL is a compiler-defined macro that tells headers whether occ is generating code that is import friendly...  e.g. against LSCRTL.DLL

as far as the __has macros,  it should be easy enough to parse them out rather than use #defines as the ocpp module already hooks occ's expression parser, as long as we can return a constant value (for example from expression_primary()) everything is fine.   Then we also could just default anything we don't know about to false instead of littering the code with a bunch of meaningless #define initializations   You can work on it if you want :smile:

chuggafan:
The _LIBCPP_DLL_VIS stuff is handled by _LIBCPP_BUILDING_LIBRARY stuff since we don't have _LIBCPP_DISABLE_VISIBILITY_ANNOTATIONS here.

The __has_ stuff I was actually wrong about having to deal with the :: things but the actual defines we're using currently for our __has macros do not line up with how they're done internally to clang AFAICT. 

__has_attribute in particular should, according to clang, deal with *ONLY* GNU-style attributes and it does it *WITHOUT* the double underscores, so theoretically our system currently does not work with at least some of the defines.

Also: I did not know that we're only generating stuff that can be imported when compiling against LSCRTL, explains a lot...

chuggafan:
Cool, the line 664 fix did work but we generate an access violation exception as well as some test errors now. So the issue was in fact the __LSCRTL_DLL define stuff

LADSoft:
yeah i just found a problem with nesting atomic-qualified structures inside other structures.   Basically doesn't allocate the space right.   I will fix it right after lunch...

LADSoft:
sorry i keep accidentally hitting the close button...

LADSoft:
i seem to have been wrong about that... it seems to handle structured atomics fine.   Not that there are any in libcxx anyway...

I'm at the point of looking in locale::id::__get() (defined in locale.cpp) where a call is made to __atomic_load_n and something is going wrong right in that area.   It is returning invalid ids so things go south...   need to take a break will get back to it later!

LADSoft:
i think i fixed the crash, it was a bug in the __atomic_add_fetch code generation...

chuggafan:
Tests are still generating stacktraces in alexcs and shorter samples than reference samples...

LADSoft:
i don't know what is up with sample3.ref.   Whenever you first load it from the git server it is broken...   but it works on appveyor???   If you fix it it doesn't get checked out again and everything is fine from then on which is why I hardly think about it.    There is a rule in .gitattributes which should be making it binary so I just don't know what is going on.

mostly occ seems to work with your changes other than that and having to download the c-testsuite...   there is a problem with stdatomic-lockfree.c though, it explicitly assumes a boolean value is going to be set to either zero or 1 but you've set the ATOMIC_XXX_LOCK_FREE constants to 2...

chuggafan:
ATOMIC_XXX_LOCK_FREE is defined to be valid as 2

https://en.cppreference.com/w/c/atomic/ATOMIC_LOCK_FREE_consts

I guess that just means the ATOMIC_*_LOCK_FREE test is completely borked...

LADSoft:
hm actually it uses atomic_lock_free() which is supposed to return a bool....   give me a minute and I will fix it to test the result against 2.

LADSoft:
hopefully i've fixed the debug build now and maybe the occil build...   I'll look at the LSCRTL build this afternoon and maybe do some other testing :smile:

LADSoft:
the lscrtl build should be fixed now;  meanwhile there are a couple of problems with actually using atomic functionality and threading functionality which I will address soon...

chuggafan:
Nice! Hopefully this means I can switch out omake to use the <thread> header...

LADSoft:
i got threads and mutexes to compile and mostly work, main problem is join() crashes right now...   I also introduced some issue somewhere that makes some of the c++ tests fail to compile.   Should be able to clean this up this weekend...

chuggafan:
Join shouldn't crash unless there's bad codegen... what are the tests you're using for mutexes?

LADSoft:
for mutexes I did something simple...   the threading test I have starts multiple threads and prints something from each thread, the output gets garbled.   So I threw in a recursive mutex along with lock_guard to clean it up...   like I say really simple...   enough to know it compiles and performs the intended operation...

LADSoft:
for the join problem, I had something similar at one point, when working on C11 threading.   There was a bug in the rtl that I fixed.  I was wondering, what is the basis you selected for the threading libray we are using?   Does it use C11?   _beginthread/_endthread and friends?   Or something else?

chuggafan:
The threading library *SHOULD* be the default Win32 thread handler, so if there's problems with it it's likely to be a problem in libcxx unless we're doing bad codegen or something weird....

LADSoft:
you are right it seems to be bad codegen.  __stdcall wasn't being applied to template class member functions (it gets applied to normal class member functions just fine).   So the fls shutdown callbacks were causing a crash in the kernel...

I'll fix it later on then with luck everything will still be compiling...

LADSoft:
i checked in the latest, but, it was kind of involved transfering stuff from one directory to another so i'm not 100% sure I transcribed all the changes properly.   If you could merge it with your PR I'll pull it and check it out again...

PR #599, opened 2021-06-22
massively improve how errors are handled

chuggafan:
This centralizes all of the errors in OCCPARSE to one file, allowing for quick, seamless debugging and referencing of errors, such as searching for error numbers, becomes instantly easy to see what error it is, what warning level it has, etc.

Technically this introduces the MSVC warning of unscoped non-class enums but I don't care enough to fix that and also, well, this works great for our purposes and we can even give ourselves an "END" so we don't have to do the `sizeof(errors) / sizeof(errors[0])` trick.


GitMensch (approved):
Nice work

LADSoft (commented):
this good.   Thanks.

chuggafan:
I just realized this PR has a 3 minute build time penalty... ocpp will *DESPERATELY* need speeding up after this because of this... maybe PCH support?

*SIGH*, that's the real problem with this: it goes into too many files.... We'll have to do performance inspections of before/after to see where hits are because of this...

LADSoft:
well i tried to use string::erase and string::replace for the heavy lifting in ppDefine.cpp...   because they will work on the buffer if it is big enough.

but it crosses my mind there might be less reallocations for string::replace if the string size was increased in advance....   don't know offhand how big an impact that would make though...

LADSoft:
im also trying to remember...  i know at one point in time builds were much faster.    it may have been the introduction of the C++ preprocessor using strings that stopped that, I don't know...   Anyway, the old version pre-c++ did everything in very large buffers, no reallocations during the preprocessing.  I do know it was faster that way...   just don't recall by how much.

chuggafan:
Now that I'm thinking about this... it could also be the compiler itself causing the slowdown because of the fact that setting the individual enum values is a very, very annoying task...

What I *COULD* do to fix this is simply drop the number from the schema and simply make ERR_ASM a duplicate of ERR_PURESTRING and see how that goes... does that fancy you for a test?

LADSoft:
you could try it and see how that goes :smile:

chuggafan:
Pushed a change that *should* do this, if you notice, I changed every error code after ERR_ASM still anyways, this is so that the file is a quick reference for every error along with their error code.

chuggafan:
Apparently this made the jump 6 minutes instead of 3 .-.

I'm now c o n f u s e d

chuggafan:
Now that I think of it, this is a good case for using VTune on my own... so the plan will be taking all 3 MSVC Release files and VTune perf test it with a PDB and find debug symbols from that... yay! getting to play with a new toy it is... will probably wait a day or two until I do that though because it's rather annoying to set up (hopefully without it forcefully BSODing me...)

LADSoft:
i forgot; i had discovered that appveyor builds during the day (us time) are worse because of server load; maybe there is really no new problem after all.    That would be nice.   I'll restart the build after it finishes to see what happens...

chuggafan:
Huh, it was an extra 6 on the gcc branch but not on any of the others... I'm now legitimately confused... is mingw's preprocessor just that bad?

*sigh*

LADSoft:
yeah.   If you want to restore this PR to its original state we can run a build tonite and see where we stand...  if it is ok we can move #600 to milestone 5 but I definitely do want to reevaluate it at some point lol.

I'm making progress on the optimizer issues.    May have the GCSE worked out tonite, not sure...   currently have to fix a bug with inserting some copies of expressions...

chuggafan:
I've returned the PR to the original state, such that ERR_ASM is just a map on ERR_PURESTRING

PR #629, opened 2021-08-02
definition of MinGW-style type defines

GitMensch:
Stuff like `_PID_T_` is queried in different projects and defined in at least the MinGW headers, so I've added the relevant ones.

Please don't feel any urge to pull that in, it is more a suggestion but also a general design issue.

Note: different projects additionally/alternatively query the defines that are found in at least glibc's types.h, like `__pid_t_defined´ it may be reasonable to check for/define both (or not...).

### DIFF
--- a/src/clibs/stdinc/sys/types.h
+++ b/src/clibs/stdinc/sys/types.h
@@ -48,7 +48,10 @@
 #        define _SIZE_T
 #        define _USING_TYPES_H_SIZE_T
 typedef unsigned size_t;
+#        ifndef	_SSIZE_T_
+#        define _SSIZE_T
 typedef int ssize_t;
+#        endif
 #    endif
 
 #    ifndef _TIME_T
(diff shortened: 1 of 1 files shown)


GitMensch:
What do you think about adding the additional defines that are found in at least glibc's types.h, like `__pid_t_defined`?
It may be reasonable to check for/define both (or not...).

LADSoft:
i guess if either is defined make sure both are defined?   Otherwise define them and also define the enclosed type?

GitMensch:
As this is a system header I think there's not the big necessity to do extensive checks for those (and I've not seen anything but system headers yet that self-define `__pid_t_defined` somewhere). If this is the single place in OrangeC that does the define (what I think it is) then possibly have the check as is now and additional define `__pid_t_defined` and similar?

I'll leave that design issue to you :-)

Issue #610, opened 2021-07-15
Creating a new format for help text and options


chuggafan:
https://github.com/LADSoft/OrangeC/commit/02453f1f84b3b4cf9b8cb7336a372b8ca699f1af#r53508135

This comment describes an idea I have for reimaginging how to format our options, so that we can categorize options for narrowing down help, we can automagically add/remove options, describe what conditions options happen under, etc.

we can do this with a macro implementation similar to the one shown in the [errorlist](https://github.com/LADSoft/OrangeC/blob/master/src/occparse/errorlist.h) and thus vastly improve option organization and centralize help text for each thing.

LADSoft:
i was actually thinking about the fact the help is spread out over different compiler configuration files tonite...

chuggafan:
I've begun some sembalance of work on this, I think in the doing of this I'm also going to pull out the FNV1A_class stuff from beintris to an outer file to allow for any class to use it, as I want to use the string as a reference to the error number without going through the expensive std::string constructor.

LADSoft:
Sure.   That sounds like a good thing to have more widely available too.

chuggafan:
While doing review of the various errors, I found ERR_FUNC_NOT_THREAD_LOCAL, error 288, which says "Functions cannot be thread local", but the error level is of "WARNING", should this be an ERROR? I can change this on my side so that there's no merge conflicts.

LADSoft:
it probably should be an error...  go ahead and change it if you would.

GitMensch:
Is this still open? @chuggafan if yes, can you edit the issue to be more explicit what the suggested changes are, please?

Issue #605, opened 2021-07-07
Support -Werror for GCC compatability


chuggafan:
https://github.com/LADSoft/OrangeC/commit/5e8957686d0cef25b42b001c9c38f9be71bebb30#r53142153

Discussion kicks off from here.

The idea for how to implement a similar system to how [gcc and clang](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) do error options, we could implement a similar -Wall -Wextra -Weverything that clang has as well giving control over experimental, guaranteed, and other options, but that's for the future.

Idea for how to do this: Go to the recently added errors file, add another parameter for the name of the error, set the name + number into a static const unordered_map (since our goal is finding, this should be fine, hopefully), and then use the existing "mark warnings as errors" functionality to set warnings as errors.

LADSoft:
we could do it in stages - the most obvious errors first and then hone in from there...

chuggafan:
The question is: will we have GCC compat for certain error names? And if so, what errors are they?

chuggafan:
Doing this while doing #610 because they're related and I think I'll use the same strategies to accomplish it. Just adding 4 macros to errorlist.h and just making it so that there's help and the GCC-compat names.

GitMensch:
Thanks @chuggafan - is there already a PR for both issues or something that you (currently/later?) work on?

chuggafan:
I'm currently working on it, there actually aren't as many shared warnings as I thought however...

GitMensch:
> I'm currently working on it

cool!

> there actually aren't as many shared warnings as I thought however...

Hm, is this because occ does not warn the stuff GCC does or because of one of those compilers combines the warnings under a single flag while the other doesn't?

In the first case it seems reasonable to create FRs for the "most reasonable warnings" missing, in the second case occ could get a new warning "group" to match the one from GCC.

chuggafan:
> Hm, is this because occ does not warn the stuff GCC does or because of one of those compilers combines the warnings under a single flag while the other doesn't?

A lot of the former, 1-2 of the latter.

A lot of it has to do with static analysis that would really slow down builds but are useful to implement for future ability to verify say, that malloc and dealloc happen.

LADSoft:
[...]

PR #635, opened 2021-09-08
Add FNV functions and more customization in error lists

chuggafan:
This PR adds the FNV functions to a namespace called "Util" (because Utils is taken by a class instead of a namespace), I also added a few support things to do GCC warning compat along with help text for some warnings.

As for actually getting the frontend to do this... I don't know enough about the utilities used for processing the input flags to modify them to accept this stuff.


chuggafan:
Adding a comment to link in #610 that this is related to but doesn't completely solve, but this should mostly fix that. This also does some work (towards the naming) for #605

LADSoft:
Sorry for the delay, I been on vacation lol.   I was thinking, we could turn 'utils' into a namespace.    I *think* that might be ok because everything in 'utils' should be static anyway.   It is probably just a matter of refactoring the header file...

LADSoft:
as far as the error processing, it would just be a matter of instantiating a CmdSwitchxxxxx class (which you could do with new).   Once the class is instantiated everything else is automagic, the first argument to the class constructor hooks it into the switch mechanisms.    But of course you have to do that before calling the switchparser...

chuggafan:
I'm going to use the already existing Werror=whatever thing and begin to use the stuff I've made.

LADSoft:
were you needing some guidance on this?

chuggafan:
I did a bunch of the work so far, just been lazy about committing atm cuz I have a lot of work this week

chuggafan:
Got back around to doing some testing on my changes, since I'm very busy these next few weeks I'll be unable to do much development, since my changes are having issues (but I think I can still add what I want), I wouldn't want the pull merged immediately once I commit my current work because it doesn't... work.

LADSoft:
are you ready to have this merged?

chuggafan:
No, the actual feature doesn't work and I've had a lot of other stuff on my mind, so I haven't had the time, if you want to figure out why 1. This decreases the compile speed enough to put us over the 1:30 mark and/or 2. why this change is not working it'd be fine... but I have no idea what's wrong with it atm and I likely won't have the time this week or next week to figure it out.

chuggafan:
Got around to testing just now, turns out the issue was really, really, reallllllllllllly simple and easy to overlook: miss-spelled a macro.

GitMensch:
If I understood that correctly it is ready to be merged in (so stays in and possibly is updated), but will need a later follow-up PR to be finished. Correct? If this is a WIP-PR please mark it that way.

chuggafan:
Technically everything that is needed for this is completed, the only things that need to be added/fixed-up are having help menu stuffs, but I'm not sure we want that and it can go into its own PR block. As-is this is perfectly fine for accomplishing what *I* want re: getting less errors when doing the libcxx tests.

LADSoft:
It looks like it probably won't mess with the build times appreciably, are you ready for me to review and merge it into the main line?

chuggafan:
Absolutely, only things to work on would be theoretical help stuff but that would likely require a rewrite of the help section itself even further to get it to a nice area.

chuggafan:
Made a few fixups so the analyzer stops compllaining.

LADSoft:
I was about to look at merging this...   but apparently the build is failing in a big way right now.    It was ok the other day except for the debug build which timed out as usual...

chuggafan:
Ye, the issue was that I wanted to apply [[noreturn]] on some functions, but OCC is unique in erroring on it if a noreturn function has a return value, I'm debating whether this should be downgraded to a warning or I should work around it.

Pros for downgrading to a warning:
- Allows marking the contract of a function being used as a function pointer, like in this case.
Cons for downgrading to a warning:
- Suspiciously marked noreturn functions are strange

chuggafan:
I've decided to do a temp workaround for this [[noreturn]] error as a demonstration that other than that everything is good, leaving the conundrum above to be shoved aside for later consideration (I should probably make an issue so we can further discuss this outside of a pull request, but I have other stuff I'm doing today....)

As for other stuff such as me moving to VS2022: meh, seems ok to do so since it still supports the VS2019 toolset, only problem could be that VS2022 is IIRC supposed to get an ABI break via vNext on the microsoft STL but other people would know far more than me about that...

chuggafan:
Gonna be honest and say that I am consistently forgetting the small things with these changes somehow... so I'll probably squash the past ~5ish commits into one then force push sometime later tonight because I should have more time to think then

LADSoft:
thanks... will make it slightly easier to review.   I'll make an issue so we don't forget the [[noreturn]] problem

chuggafan:
Got around to it just now, so all the relevant analyzer fixes are squashed

PR #658, opened 2021-12-17
add the _mm_pause intrinsic

### LINKED ISSUE #657: Adding __mm_pause to the backend
The reason I'm creating this as an issue is that even with the documentation touching the x64.adl system is... interesting...

I specifically know that currently the `pause` instruction is unsupported by the infrastructure, and what I'd like to do is use this to do spin-wait loops in an upcoming job-server patch ever so-slightly more efficient. But I can't add the intrinsic to the code if the instruction quite literally does not exist, after all.

I *THINK* something like `<Opcode Name="pause" mand="" op="0xf3:8 0x90:8" Class="sse2"/>` is *close* to what I want based off of the opcode documentation [here](https://www.felixcloutier.com/x86/pause) but I'm not quite sure

chuggafan:
This also adds in the pause instruction into the assembler, as
previously this was unavailable to the assembler

Also: I am *really* uncertain about the veracity of the pause instruction, I'm not sure if the mand is what should use the `0xf3:8` or that `op` contains the `0xf3:8`, since mand seems to be a prefix modifier(?)

When this gets pulled this should fix #657


LADSoft:
` mand` is only used in addressing modes, e.g. in operands.   Since the instruction has no operands setting `mand` wouldn't do anything, so just setting it in the `op` field should be fine.   I'll check it out later on...

Issue #659, opened 2021-12-30
Question re: design of OMAKE


chuggafan:
I'm trying to do a big redesign of the multithreading for OMAKE, and I keep seeing the processIDSemaphore that exists in the OS system, I know this is used intra-process for the executing process ID list, but I don't see where this gets used outside of being a (somewhat jank) inter-process mutex for an intra-thread item, is there something I'm missing with the design of this or am I able to swap this over to an internal mutex for better compatibility with GNU Make?

LADSoft:
omake spawns itself; there is a named sem shared between all instances of omake that prevent it from starting too many non-omake subprocesses.   So yeah it is used internally, but at the same time it is not used internally because of the multiple running omake instances.

chuggafan:
I meant the processID semaphore, called `processIdSem`, which only seems to guard the static `std::set` of process IDs, this isn't an interprocess communication as far as I can tell, so am I OK to just make `processIdSem` a mutex since it only has a max of 1 user at a time?

LADSoft:
the environment variable .OMAKESEM holds the semaphore name for a recursive set of calls to omake.   It is initialized by the top level omake and passed through the environment to recursive calls to omake.   Line 246 of os.cpp opens a named semaphore with that name (which is shared by the recursive set of calls).   So I think the semaphore is shared by different invocations of omake, and turning it into a local mutex wouldn't work so well.

chuggafan:
Different question this time: I see that the print directory stuffs is handled in the JobInit function, and I can kind of understand that there's a file that's temporarily created (somehow). [_tempnam](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=msvc-170) is used, so presumably there's a blank directory this temporary file is located? 
Why is this in the JobInit section, what can I do to move it out of the JobInit function, and how can I stop it's reliance on the OMAKE semaphore server name and have it be called something else so that I can begin a move to a more GNU make compatible version?

LADSoft:
tempnam uses the path found in the TMP environment variable if there is one, or else it uses the first argument to tempnam.   In this case if TMP doesn't exist it would use the current directory.   But it makes up a random file name...   I think @GitMensch pointed out there is a better alternative than using tempnam but I haven't looked into that.

[...]

Issue #656, opened 2021-12-16
Separate files for OMAKE's OS for linux and windows


chuggafan:
This is a buildsystem issue, but basically what I want to do is move the OS.cpp into two files:

`os_specific\unix\posix_os.cpp`
and
`os_specific\windows\windows_os.cpp`

This is so that os.cpp is no longer an `ifdef` soup and because I want to try to add a ["jobserver"](https://www.gnu.org/software/make/manual/html_node/Job-Slots.html) implementation from GNU Make for the ability to do semaphores and the like on Linux.

I don't know how much this could complicate the current already extremely complex build system though...

LADSoft:
the easy way is to use conditionals in the makefile, to add one of the new files to the CPP_DEPENDENCIES:= variable 

```
ifeq "$(COMPILER)" "gcc-linux"
CPP_DEPENDENCIES:= $(CPP_DEPENDENCIES) os_specific\unix\posix_os.cpp
else
CPP_DEPENDENCIES:= $(CPP_DEPENDENCIES) os_specific\windows\windows_os.cpp
endif
```

but there might be a bug in the builder where it doesn't strip the directory off when creating an object file name...   easy enough to fix though I'm sure...

another possibility is create a third file os.cpp which conditionally includes one of the other two...

chuggafan:
Tried this with a project for OMAKE that I'm working on to make OMAKE accept the recursive makefile calls GNU Make does so that if one calls the other we can pass around the maximum number of threads each can take

Currently having an error like this when I modify the makefile:

```
Error C:\OrangeC\src\/ms.mak(112): No rule to make target 'Win_Jobserver.obj'
Error C:\OrangeC\src\/ms.mak(117): No rule to make target 'omake.lib'
```
The modified makefile looks like:

```make

include ../pathext2.mak

NAME=omake
MAIN_FILE=MakeMain.cpp
INCLUDES=..$(PATHEXT2)util
ifeq "$(COMPILER)" "gcc-linux"
CPP_DEPENDENCIES=$(wildcard *.cpp) os_specific/Linux/Linux_Jobserver.cpp
else
CPP_DEPENDENCIES=$(wildcard *.cpp) os_specific/windows/Win_Jobserver.cpp
endif
LIB_DEPENDENCIES=util
H_FILES=$(wildcard *.h)

include ../redirect.mak

DISTRIBUTE: copyexe
```

I have *NO* idea why it says there's no rule to make it unless:
```make
[...]

Issue #56, opened 2018-04-11
omake is not GNU make compatible, raises Assertion


GitMensch:
I guess simple ones work but would recommend to reword the reference to GNU make.

Just a small sample that doesn't work (actually raises a CRT error): [Makefile](https://github.com/LADSoft/OrangeC/files/1900739/Makefile.txt)

and running:

~~~
omake clean
Error Makefile(14): Command without a rule
Error Makefile(18): Command without a rule
Error Makefile(37): Command without a rule
Error Makefile(38): Command without a rule
Error Makefile(59): Command without a rule
Error Makefile(60): Command without a rule
Error Makefile(61): Command without a rule
Error Makefile(62): Command without a rule
Error Makefile(63): Command without a rule
~~~

and

~~~
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Debug Assertion Failed!

Program: D:\dev\OrangeC\src\Debug\omake.exe
c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.13.26128\include\xstring
Line: 3148

Expression: string subscript out of range

For information on how your program can cause an assertion
failure, see the Visual C++ documentation on asserts.
~~~

LADSoft:
I reviewed the makefile and it seems buggy that omake wouldn't accept it...   so I'll fix that.   But am I going to take your advice and reword the reference as well since I don't have an adequate means to test real make files.   Probably do this over the weekend.

GitMensch:
Attaching another one which uses GNU make extensions (at least the wildcard).
It would be nice to be able to process this with omake, too.

I didn't checked all but the build-in targets seem to be not recognized, too.
In my opinion the first thing to do there is to recognize them and *warn* about "currently ignored" where the actual implementation needs more time (this way omake can at least process these files).
For a list of these targets see:

* [GNU make, special targets](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html)
* [BSD target attributes](https://www.freebsd.org/doc/en_US.ISO8859-1/books/pmake/targetattr.html)

LADSoft:
I am able to solve the first two issues:   the main one was a vagary with how '\t' is processed by omake.

Thanks for the documentation.   I checked it and the only unimplemented special targets are as follows:

.LOW_RESOLUTION_TIME
.NOTPARALLEL
.ONESHELL
.POSIX

They look easy to implement so I will implement and check the others to make sure they still work as intended.

[...]

PR #660, opened 2021-12-31
Use a JobServer compatible with GNU Make

chuggafan:
This PR is currently incomplete, and as thus is left as a draft, I am making this so that my work can be rightfully critiqued so that it can fit in.

Changes I have done but haven't discussed:

I have begun working on abolishing the mandated timeout scheme, I know this could (theoretically) break builds again but we should be fixing that we're getting stuck instead of timing out and admitting defeat

namespace OMAKE

I haven't put a license on the code header, but IDRC what the license for the code I create is, we can place it as GPLv3 like the rest if you so choose

I am using std::thread instead of doing CloseHandle on _beginthreadex to make sure that cleanup happens, as well as type safety

I am making SPAWNER use std::shared_future so that when we join() all the threads at the end of this we can be happy

I renamed WriteConsole, because Visual Studio was complaining about it and having build issues because the name overlapped with the windows.h define for "WriteConsole", causing build issues at some point while I was working on this

Again, this is incomplete for now and it isn't *fully tested*, a lot of code still needs to move around for me to feel that this is acceptable for even *beginning proper testing* other than a "This Proof of Concept works in some vague manner"

#659 is related to this
#656 is currently sidestepped by having a "master" file that will include the other two types depending on what goes on
Forgot to mention that this will likely be a "final boss" for #56 as well


LADSoft (commented):
I like this whole jobserver thing.    TBH I've only been exposed to these kinds of indirections within say the last five years or so but I like them!

chuggafan:
I forgot to mention RE: this PR, this PR will definitely be failing until we can get the threading passes done, which is why I was inquiring about debugging it before. This is partially because the plan is to get out of using makeshift std::futures and move into using std::future and the like properly. I'll still be doing a *ton* of work on this, but I don't expect to get particularly far until I can properly test *just* omake and not also occ in the process

chuggafan:
Apparently part of the problem is in my semaphore implementation and *default constructors*. I have *NO IDEA* what is going on with that since I put in some pretty clear checks to prevent handles being closed when there's nothing to do.... *sighhhhhhhh* this is going to be difficult to debug...

chuggafan:
I'm even more confused now: The destructor is either being called twice or *something else* is happening, because when I assign the (NON-TEMPORARY VALUE) the destructor gets called, then gets called again,
```
204
        copy orc.exe ..\bin
        1 file(s) copied.

Error makefile(82): commands returned error code -2147483645

Error makefile(85): commands returned error code -2147483645

Error makefile(87): commands returned error code -2147483645

Error makefile(89): commands returned error code -2147483645

Error makefile(93): commands returned error code -2147483645
204
```
the dual "204"s are evidence of this happening, I *THINK* this *very well* could be an issue with MSVC

GitMensch:
Where is the destructor you reference here and where did you added the debug output of the internal id (I _guess_ that's what the 204 is)?

chuggafan:
I got it working by *removing my software break point* apparently? I'm *REALLY CONFUSED* right now... I'm pushing the code and asking that I *truly* need help to figure out wtf is going on now...

LADSoft:
omake calls itself recursively in our build scripts...   each recursive call would call the destructor when it exits.   So you could easily get multiple destructor calls, when running one of our scripts.

as far as stray software breakpoints...   I'm not 100% sure what MSVC does with them if you manage to get one into the code.    But I think OCC just crashes if you are running outside a debugger.   I tried to fix that once but I guess the fix didn't take... I haven't thought about it in a long time.

chuggafan:
Yes, I know, but this is before OMAKE gets called, they were crashing before the constructor was getting called because of the BreakPoint call, the problem extends from the parent OMAKE doing the following:
```
Default constructor
Named and initialize constructor
200, 0
        copy orc.exe ..\bin
        1 file(s) copied.

Error makefile(82): commands returned error code -2147483645

Error makefile(85): commands returned error code -2147483645

Error makefile(87): commands returned error code -2147483645

Error makefile(89): commands returned error code -2147483645

Error makefile(93): commands returned error code -2147483645
200, 0
```

This is what it looks like when I place the software break point *above* the call to DefaultConstructor, making it *EXTREMELY* clear that the issue at hand is that somehow there's a copy of this floating around and possibly closing the handle when it shouldn't be, because I moved it.... maybe that's the issue.... but I can't think of why....

I'm going to try deleting the copy constructor for now (since we're no longer using this as a standalone thing and more like a thin shim for a named semaphore)

chuggafan:
Deleting the copy constructor revealed the truth: even with the std::moves we're doing *copies*, this is a real problem it seems and it seems that I have to either create a custom copy function or just do move assignments for now...

I'm going to do the former starting tomorrow and see how this goes, if this all goes well and works I will switch over to the new threading and test that and see if it all performs together, after that, the plan is to wait until there's a working std threading base to use this on so that no shenanigans happen here...

LADSoft:
hm, I wasn't aware of any current issues with copy constructors being called when move constructors should be called...   I know in the past there have been such issues and they did cause things to crash though...   do you want me to change focus and look at the threading tests for a while so I can try to discover what is going on?

LADSoft:
I couldn't figure out what part of the code was triggering problems so i tried this:

```
#include "Semaphores.h"
int main()
{
	Semaphore sem1, sem2;
	sem1 = std::move(sem2);
	sem1 = sem2;
}
```

and you are right it does use the copy assignment operator.    I did some digging and at least with the current code base that seems correct according to the other compilers; your Semaphore implementation seems to not allow for automatic creation of a move assignment operator and the copy assignment operator is used instead...  (I can never remember the rules for this sigh).

chuggafan:
Yhea, my plans are to create a custom copy assignment operator that will open a reference to the semaphore before closing it prematurely, and simply drop the non-named-semaphore support (or internally name the semaphore, maybe with a similar random generation scheme that's currently going on)

chuggafan:
I believe this version works, I tested it with DebugCompiles, but build issues with occ itself hamstrung it. I thus am likely in a few weeks time to come back to this to implement the new threading system as its testbed. We should also deal nicely with the infinite looping issue, with surgical identification of where and most importantly, *why*, so that this program can always run correctly

chuggafan:
*ap*arrently we don't have getpid in our standard library. We can just shim this with a thin callout to GetProcessId

GitMensch:
Please add the wrapper for `getpid()` directly in the RTS, not only in this PR, and ideally also `getppid()` along with `pid_t` and the define `__pid_t_defined` into `unistd.h`. Thanks!

chuggafan:
I meant it that way, we can make the getpid function do a thin call to GetProcessId, IDK if we want to support DOS anymore so I'll leave it in there for windows alone since I'm not sure what *its* getpid function would look like

chuggafan:
Thinking some more: windows currently does not implement getppid because getppid is a significantly harder task to perform on windows as best as I can tell, so ideally all we need currently is getpid, @GitMensch is this good with you?

GitMensch:
Nah...

But of course I'm fine with whatever gets in ;-)
Implementation: https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4

chuggafan:
Turns out we already had pid_t and getpid, whoops, trying to figure out where windows defines it now so we can also import it in the same place...

chuggafan:
Turns out I'm way dumber than I thought I was, process.h had it and was never included anywhere except on windows .-.

chuggafan:
Current issue with compilation seems to be OCC not handling overloaded `.` and `->` operators correctly for std::shared_ptr and std::unique_ptr

What *fun* that is

LADSoft:
AFAIK you can't overload the '.' operator.   As far as the -> operator there doesn''t seem to be anything fundamentally wrong, as the following code snippet compiles and prints the desired results.   Can you point me to something that doesn't work?

```
#include <memory>
struct aa
{
	aa( int r,int s, int t) : c(r), d(s), e(t) { }
	int c, d, e;

};

main()
{
	auto t = std::make_shared<aa>(aa(1,2,3));
	printf("%d %d %d\n", t->c, t->d, t->e);
	auto t1 = std::make_unique<aa>(aa(4,5,6));
	printf("%d %d %d", t1->c, t1->d, t1->e);
}

```

chuggafan:
```
Error( 60)    Eval.cpp(429):  Structure required to left of '.'

Error( 59)    Eval.cpp(431):  Pointer to structure required to left of '->'
```
This is an error I'm currently getting

chuggafan:
I've at least tracked down that it's happening in expression_member because of the fact that it's triggering specifically on the lex GetCommands stuff... no idea why this is going on at this time.

GitMensch:
To close this loose end - we don't only have getpid but also getppid https://github.com/LADSoft/OrangeC/blob/master/src/clibs/platform/win32/rtl/getpid.c  - the only thing that possibly should be added in this implementation is error handling, see https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4.

LADSoft:
@chuggafan, I'm making progress, the problem isn't the member operators...   the typedef on line 115 of rule.h is being processed incorrectly and is returning an iterator to a list of ints...    simple attempts to reproduce the issue don't result in problems so I have to debug it in situ...  i'll try to get it fixed tomorrow...

chuggafan:
Thanks, after that I'll be looking into the 2nd part of this PR:

Moving to use std::thread and std::future, as well as beginning to find out just why we get stuck in infinite loops sometimes....

LADSoft:
i was able to fix the first issue relatively easily, however, there are other issues with spawner.cpp I'm looking at now...

LADSoft:
@chuggafan I fixed spawner.cpp and I fixed some other stuff, the current version of occ compiles all your code.

chuggafan:
Getting `Error: Undefined External 'std::__1::thread::thread( const thread&)' in module Spawner.cpp` when doing the changes here now, no idea why it's complaining about a deleted copy constructor....

chuggafan:
Might be some *other* reason though, I'll probably throw up the new changes I've made (such as change the thread number when -J is unspecified to the hardware_concurrency limit)

LADSoft:
I'm still having problems with that, apparently...  yesterday your initialization of the SpawnerTracker with open braces had similar problems...   whenever you put up the new version let me know and I'll look into it :)

chuggafan:
Pushed the latest changes, for some reason it wasn't accepting me do a regular push (git being wrong and weird), so I fixed that as well while doing this...

LADSoft:
I've been having troubles with the windows.h file...   for some reason sometimes it gets 'stuck' and I can't clear it out of my status window.   It can be a hassle to go back and forth between different branches...

anyway will try to look at this tomorrow evening.

chuggafan:
I think it'd have to do with line-endings, currently it seems to actually just be an access violation now... but not on my machine! HEHEHEHEHEHEHEHE

chuggafan:
So it appears that the compile error *only happens* when compiling in debug mode. That's *FUN*

LADSoft:
hehehe...   yeah i dunno why it isn't a problem in the release version as well, the reference to the thread copy constructor is still made there.   Only thing I can think of without a lot of analysis is that it is getting optimized away at link time?

apparently there is some problem though, with using the move constructor in the auto-generated move constructor for Spawner::SpawnerTracker().   It is preferring the copy constructor....   thought I had fixed that....   back to the drawing board I guess...   made me think though, maybe I need to put a comment next to each function reference in the assembly language file, indicating what the unmangled function is lol...   I'm kinda used to looking at mangled names but not so much for anyone else and anyway if we ever go to microsoft name mangling I'd be as lost as anyone lol...

Meanwhile I've been noticing that even when I compile your new version of omake with MSVC, it just kinda hangs when I run it....   is it still a WIP?

LADSoft:
I fixed it.   Incidentally that fixed 30 of the container tests so I get to feel like I'm making progress there too :smile:

chuggafan:
Yhea, I've noticed that it hangs when running without any multithreading at the moment, I'm going to try to investigate the reason for this, but I heavily suspect it's because of the JobServer being taken and untaken and still being passed along when trying to perform the issues. Next up on my list will be investigating where the access violation is...

chuggafan:
Using a rebase seems to completely screw over my git history these days... I dunno what's up with that now...

Anyways, on simple tests on my machine this new jobserver usage seems to have fixed the issue, the issue was that my design (incorrectly) did not assume that it had a guaranteed free job when it is created, which it by design does.

Another issue that needs to be batted up and I'll save it for tomorrow: access violation (eughck), I'm going to do a pass-over with AddressSanitizer while running the thing and see just how far I'll get when doing so...

chuggafan:
```
Access Violation:(C:\orangec\temp\omake.exe)
CS:EIP 0023:004DBFB4  SS:ESP 002B:0272F43C
EAX: 0272F800  EBX: 022804DC  ECX: 696C5C73  EDX: 696C5C73  flags: 00010202
EBP: 0272F800  ESI: 696C5C63  EDI: 022804D8
 DS:     002B   ES:     002B   FS:     0053   GS:     002B


CS:EIP  8B 09 8B 00 50 51 E8 D7 63 FD FF 59 59 23 C0 0F
```

I'm currently trying to debug this by comparing this location against the map file, I don't know which of these values I'm supposed to be taking in order to compare against the map file, CS:EIP seems to be the logical choice but I'm not sure how to correlate that to a map-file value...

LADSoft:
the map file is a little hokey because the virtual functions are listed as offsets into a section, rather than as pure function addresses.    Could use some improvement there...  and also I'm not sure the functions aren't mangled in the map file which would be a further difficulty...   although maybe I fixed that one lol.


anyway yeah you need the CS:EIP.    But that usually doesn't give you an address directly in the map file because it won't be the start of a function....   what I usually do is run it in the debugger and find the start of the function so I can have an address that will be in the map file.

Anyway, that aside...   if the function is not virtual then you can use the address more or less directly to interpolate the publics by address table.   If it is virtual (which you can figure out by looking at the address of the section that combines all the virtual functions) then you have to look at the offsets for each virtual section until you find the one that contains your address.    Those offsets are zero based, from the section base address...   which is usually 0x401000.   

But what i often do is take the address of the function that I got out of the debugger, subtract 0x401000, and use grep to find it in the map file...

but this whole thing could be cleaned up a bit I guess...   I'll write an issue to make sure it gets documented/cleaned up later...

chuggafan:
Cool, thanks, the boiled down essence is:
Take the address from CS:EIP, subtract the base virtual address (generally 0x401000), then grep the first few digits in the file...
What I *SHOULD* do is take the functions we have an address + size for and automate the comparison for "What function we're in", so that if I dump an address we can get the correct function thing, but that seems like too much effort, I am now debugging why we're having crashes, and it is still unknown to me why Clang makes us have OOM errors...

LADSoft:
that will work for virtual functions.    For nonvirtual functions you don't need to subtract the 0x401000 because they are listed properly in the publics table :smile:

chuggafan:
Thanks! I've already tracked down a general idea of where it is, just need to demangle the symbol from the mapfile now, so I'm waiting on a compile to finish up there so I can just onm the information out of it.

chuggafan:
```
vsc@std@__1@#__map_value_compare$px#basic_string$c#char_traits$c~#allocator$c~~#__value_type$px#basic_string$c#char_traits$c~#allocator$c~~#unique_ptr$8Variable#default_delete$n0~~~22@VariableContainer@vlt4bool?0?~@$bcall$xqrx#__value_type$px#basic_string$c#char_traits$c~#allocator$c~~#unique_ptr$n0#default_delete$n0~~~rxpx#basic_string$c#char_traits$c~#allocator$c~~
```

This is where the current issue lies, I think what should be done in order to alleviate the issue since onm was *not* working for me is to create a specific executable that can take an input for a mangled name and output its demangled version with text. As far as I can tell it has to do with the constructor call of the VariableContainer since `bcall` is involved in this chain...

LADSoft:
what is the specific problem with onm?   it doesn't get tested often lol!

but if you want to give it a shot the name demangler used by the linker is in objlib\symbol.cpp.    While not exactly the same file as in the compiler, I think it has been kept up-to-date (the core routines get cut-and-pasted).  So you would include ObjSymbol.h and use the library objlib.l  You can either call unmangle() directly (although it may not be prototyped for view outside the module) or instantiate an ObjSymbol object and call its GetDisplayName() function...

chuggafan:
I tried using onm on Variable.o in the omake/obj/occ directory, got this output:
```
C:\OrangeC\src\omake\obj\occ>onm Variable.o
onm (OrangeC) Version 6.0.51.1
Copyright (C) LADSoft 2006-2021
Syntax Error in line 0
```

I will look into your solution, I think the way I'll design it is that you can take an input file and get a demangled output... for now I'll stick to just making a test program in Visual Studio that just demangles it... I think that will work for now.

GitMensch:
> I think the way I'll design it is that you can take an input file and get a demangled output... for now I'll stick to just making a test program in Visual Studio that just demangles it...

Sounds reasonable for now and good for later - maybe "reference in new issue" so it is easier to track and reference in the ChangeLog?

chuggafan:
I attempted to get the display name by doing:
```cpp
#include <iostream>
#include "../objlib/ObjSymbol.h"
int main()
{
    std::cout << ObjSymbol(
                     "vsc@std@__1@#__map_value_compare$px#basic_string$c#char_traits$c~#allocator$c~~#__value_type$px#basic_string$"
                     "c#char_traits$c~#allocator$c~~#unique_ptr$8Variable#default_delete$n0~~~22@VariableContainer@vlt4bool?0?~@$"
                     "bcall$xqrx#__value_type$px#basic_string$c#char_traits$c~#allocator$c~~#unique_ptr$n0#default_delete$n0~~~"
                     "rxpx#basic_string$c#char_traits$c~#allocator$c~~",
                     ObjSymbol::eAuto, 0)
                     .GetDisplayName()
              << std::endl;
}
```
But all I got was the same junk that was input, I think the demangler is borked for the symbol.... I am also pretty sure the old demangler even in occparse is out of date since the `allow __stdcall to work with template functions` commit with ID b380d7e6ef070559056beeedd1e4a9a58dfcfd04 which changed when certain things were stdcall, but I have no clue if that would change it or not...

LADSoft:
strip off the 'vsc@' as that isn't part of the mangled name (it is used as a linker flag)

chuggafan:
That fixed it, it seems it has to do with the allocations...

```
std::__1::__map_value_compare<basic_string<char, char_traits<char>, allocator<char>> const *, __value_type<basic_string<char, char_traits<char>, allocator<char>> const *, unique_ptr<Variable, default_delete<Variable>>>, VariableContainer::vlt, bool=0>::operator ()( const __value_type<basic_string<char, char_traits<char>, allocator<char>> const *, unique_ptr<Variable, default_delete<Variable>>>&,  const basic_string<char, char_traits<char>, allocator<char>> const *&) const
```
```cpp
    _LIBCPP_INLINE_VISIBILITY
    bool operator()(const _CP& __x, const _CP& __y) const
        {return static_cast<const _Compare&>(*this)(__x.__get_value().first, __y.__get_value().first);}
    _LIBCPP_INLINE_VISIBILITY
    bool operator()(const _CP& __x, const _Key& __y) const
        {return static_cast<const _Compare&>(*this)(__x.__get_value().first, __y);}
    _LIBCPP_INLINE_VISIBILITY
    bool operator()(const _Key& __x, const _CP& __y) const
        {return static_cast<const _Compare&>(*this)(__x, __y.__get_value().first);}
```
I'm thinking the VariableContainer should switch from `const std::string*` to `std::string`, does this make sense to you

LADSoft:
I think i was trying to be clever and save some memory or something, but, I don't see any reason the map key has to be const string*.   Like you say you could make it just a string and I think that will even improve the code.    Won't have the interdependency between two objects that way...    Gets rid of that ugly vlt thing too!    

There may be other opportunities to do similar elsewhere,  not sure.   I seem to remember doing similar to the vlt in other places...

chuggafan:
Yep, that fixed the first issue I think, the next up is:
`std::__1::basic_string<char, char_traits<char>, allocator<char>>::__is_long() const` for some... random ass reason? I really wish this was giving me a stacktrace because then I could figure out wtf was going on.

LADSoft:
you could try compiling for debug....   if you can get it to fail you might get a stack trace that way.     It is a little cleaner too as it doesn't do nearly as much inlining when you are compiling for debug...

chuggafan:
DEBUGCompiles.bat isn't having the stack-track generate for omake for some insane reason that I cannot comprehend at the moment.

LADSoft:
i think depending on which build you do lsdbghelper.dll may not be copied to \orangec\bin.    Executables load that library to do a stack trace, if it isn't there they just go on without.

if you don;t have it build the RTL then snag the file from \orangec\src\clibs\object\startup

chuggafan:
Ah, I think I know what the problem is then:
lsdgbhelper.dll is not created yet, causing the stack trace to not be available but the rest of the info to be available, would that make sense?
I think it makes perfect sense for how things are going that that is the case.... is there a way I can copy it to a temp location and reference it from the temp location or is that not a thing?

chuggafan:
It appears that my changes seem to be working incorrectly, I'm assuming this has something to do with the job server name improperly getting passed on, will investigate this as I definitely should not be seeing 400 instances of occparse being spawned when I only asked for 16.

LADSoft (approved):

LADSoft:
yeah that is the issue with lsdbghelper.dll.    The programs use LoadLibrary() to load it so I guess you could put it anywhere on the path?


OMAKESEM was initialized with:

```
new Variable(".OMAKESEM" , name, Variable::f_recursive, Variable::o_environ);
```
 where the o_envion means copy it to the environment when spawning new programs...   in general that is how the 'export' keyword in omake works to force a variable to be exported...

chuggafan:
Yhea, as it turns out though ".OMAKESEM" is not getting passed through on the environment variable stack, and I don't know why, causing a huge number of processese to be generated, no idea why it isn't, but all I can say is that when I added the check in the LoadEnvironment system it did *not* work and I don't know why other than SetEnvironment not being called for OMAKESEM

LADSoft:
i misspoke, it is actually the v->setExport(true) that makes it export to the environment...   that is still intact, as is the push of the variable into the variable container so that isn't the problem.

But the problem is probably here, somewhere around line 500 of MakeMain.cpp:

```
    OS::InitJobServer(); 

    bool done = false;
    Eval::SetWarnings(warnUndef.GetValue());
    while (!done && !Eval::GetErrCount())
    {
        VariableContainer::Instance()->Clear();
        RuleContainer::Instance()->Clear();
        Include::Instance()->Clear();
        Maker::ClearFirstGoal();
```
it is initializing the variable in the InitJobSem() call, but the first line of the while loop erases the variable list so the initialized value gets lost...

chuggafan:
That might do it.... I think I need to redo my design then to account for that...
I'm thinking that my InitJobServer call will only do the initialization once per run, and just re-initialize the ".OMAKESEM" variable every time...

chuggafan:
Got it all working again! Until we compile ourselves, then we have an issue with `MakeMain.cpp(vsc@std@__1@#__libcpp_atomic_refcount_decrement$l~$qrl`, I'm not using the latest master though so I don't know why `__libcpp_atomic_refcount_decrement` is failing, possibly the atomics are borked again?

```cpp
template <class _Tp>
inline _LIBCPP_INLINE_VISIBILITY _Tp
__libcpp_atomic_refcount_decrement(_Tp& __t) _NOEXCEPT
{
#if defined(_LIBCPP_HAS_BUILTIN_ATOMIC_SUPPORT) && !defined(_LIBCPP_HAS_NO_THREADS)
    return __atomic_add_fetch(&__t, -1, __ATOMIC_ACQ_REL);
#else
    return __t -= 1;
#endif
}
```

This is the offending function afaict, attempts to get the stack trace by placing lsdebughelper.dll into a safe location of the PATH have failed me, as well as attempts by placing it in the same location...

LADSoft:
glad you got the basic functionality working :smile:

maybe there is still a bug in the atomics...   last time they worked until I made some adjustments elsewhere and then bugs in the code generation took over...   I did run the atomic tests the other day and they all passed for what its worth...

I don't know why the stack trace isn't working.   I'll look at it tomorrow...   I will say without my regular testing with OCIDE there is a possibility I broke the debug info...

chuggafan:
It could be a unique interaction with the shared_ptrs and unique_ptrs as well, to be fair.

chuggafan:
A redo of the compile leads me to:
`      File: Rule.cpp(vsc@std@__1@#__compressed_pair$ui#allocator$#__list_node$#unique_ptr$4Rule#default_delete$n0~~pv~~~@$bdtr$qv) addr=095ce2 size=001c` on the first instruction
This makes me believe strongly (even moreso than the fact that that thing only appeared in this already) that this is a unique_ptr interaction with compressed pairs. I think a huge overhaul for how we're handling this should be done and an analysis for whether any of these pointers are "owned" or not...

chuggafan:
🤔 Some testing is having OMAKE spit out:
```
C:\OrangeC\src>c:\orangec\temp\omake /DCOMPILER=OCC clean -j:8         
The job server name is: OMAKE113164Error C:\OrangeC\src\treetop.mak(268): No rule to make target '☺i☺↑☻I☻╪Σ↓ÇΣ↓┤▼F☻ê/C☻É▬C☻╚♣M☻D╪ÿYGD╪↑'
```
as an error, and I have absolutely no idea what could be the cause of this, even as a theoretical, something's up with my code and I don't know what exactly it is...

GitMensch:
That's obviously some wrong data at the pointer that should contain the target name - either a memcpy or similar has gone wrong or the pointer is not correct. The common way to check this would be to start the debugger and set hardware watchpoints on both.

chuggafan:
Yhea... the really fun part is that it depends on how it's compiled, so when I do it like the appveyor is done it happens, but when I compile it using DEBUGcompiles.bat it's just doing the access violation.

*FUN*

LADSoft:
don't know about the 'junk' unless it is references to other DLLs...   the stack trace is a little unreliable in release mode though because it relies on using EBP and EBP isn't always updated unless you use /C-E compiler switch.

There were several bugs in the helper dll and RTL...   I think I've cleaned it up so it will work better at least in debug mode...   you do need to rebuild the RTL for this as one of the bugs was that it wasn't being compiled with EBP enabled...

LADSoft:
i figured out about the 'junk' - when you compile for release (without /C-E) EBP can have indeterminate values based on the program flow.   If it just happens to point into the stack it may or may not point at a valid stack frame, but the assumption has been it will...    I'm fixing this in the next push, to not allow stack frames that have IP addresses outside our code space...   but that also means you can't get a stack from from an external dll like kernel32.dll.   Which I think is ok since EBP is indeterminate within that dll as well...

chuggafan:
That might be fine, the reason I was saying it was junk was that although it was seemingly obvious that it was a location within our codespace (the virtual functions were within a specific file's range), there was no function that seemingly mapped to it, at least with what I could test.

I'm likely to continue testing later today and see where that crash is now...

chuggafan:
Whelp, your fixes did work, and it does trace back to where I thought it did:

```
Stack trace:
                        46942a: std::__1::__libcpp_atomic_refcount_decrement<long>(long&) + 0xc  module: makemain.cpp, line: 3386
                        46837d: std::__1::__shared_count::__release_shared() + 0xf  module: makemain.cpp, line: 3439
                        466588: std::__1::__shared_weak_count::__release_shared() + 0xc  module: makemain.cpp, line: 3482
                        464904: std::__1::shared_ptr<RuleContainer>::~shared_ptr() + 0x18  module: makemain.cpp, line: 4207
                        412a29: Maker::Dependencies( const basic_string<char, char_traits<char>, allocator<char>>&,  const basic_string<char, char_traits<char>, allocator<char>>&, Time&, bool, basic_string<char, char_traits<char>, 
allocator<char>>, int) + 0x10b4  module: maker.cpp, line: 302
                        411809: Maker::CreateDependencyTree() + 0x3b7  module: maker.cpp, line: 812
                        404a17: MakeMain::Run(int, char**) + 0x1e40  module: makemain.cpp, line: 668
                        40105c: main + 0x42  module: makemain.cpp, line: 100
                        43b447: __startup + 0x1bb
```
It doesn't go off when doing the DEBUGcompiles.bat unless I add the location to PATH and copy it into a temp location, which should probably be part of it as a step...
Anyways, the only problem I'm having with this stack trace is now I have to figure out from what file (makemain) I need to actually go to, since the original file (e.g. FILE directives) seem to be ineffective in this debug info... I *THINK* the 0xc is enough to be something like 2 instructions? So I'm 100% sure somehow a null pointer atomic is being deferenced here and I don't know why...

LADSoft:
yeah i noticed the files it lists are wonky... can you write an issue so I remember to look at it next week?   Hopefully the info is there and I just have to fetch it from the DB....

LADSoft:
so i don't know if you noticed but i fixed ObjSwitchInt to consider its argument optional.  I defaulted the -j switch to one in linkermain.cpp.    You could set it to zero if you want and then check for a zero when processing the switch,   use that as a flag to indicate you can substitute it with whatever default you decide on.

I also fixed #668 so that it should show you the correct file information now.   The bug was in the compiler so you have to recompile to get it.

I played with your version of omake for a while....    pretty good but needs a little work.   I could only get it to crash sometimes when trying to use it to build orangec although there were some other messy issues...   I was having trouble with the builds deleting the debug file when I was in there. so I never quite got to the point of seeing a proper crash dump.

It looks good so far...   do you need anything from me immediately?

chuggafan:
I am still having issues with the thing I said earlier, I think it has to do with faulty atomic generation... I'll push the current code at some point tonight probably, my solution has been to have the debugging dll on the path in C:\OrangeC\temp with a copy of the dll that I got by interrupting the compile at a point... I'll look at the code you uploaded today sometime tonight, and then see if the crash is still there, if it is, then I'll probably open up OCIDE and take a look at debugging the assembly output if I can...

chuggafan:
So I've taken a look at the disassembly here:
```assembly
                 push        ebp
                 mov         ebp,esp
                 mov         eax,dword ptr [ebp+8]
                 mov         ecx,dword ptr [eax]
                 dec         ecx
                 mov         eax,dword ptr [ebp+8]
                 mov         dword ptr [eax],ecx
```
And it crashes on this last line, leading to for me to believe that the ebp+8 movement on the 2nd time causes an issue, I do not know why it would, but it does, the other thing I noticed is that the decrement is *not* locked, at least, according to DUMPBIN, so I'm not sure if it's atomic generation that's broken or if it's something else entirely...

LADSoft:
ok when uses EBP with a positive value, it is accessing arguments.   EBP+ 8 is the first argument (or the this pointer for member functions or the pointer to a structured return value for functions that return a structure by-value), EBP+0c is the second argument or the first argument or the thispointer depending on what else is stacked, and so on (a long long takes two argument slots and a structure by value takes as many argument slots as it takes to store the structure).   So what this says to me is a pointer which points to an unwriteable region...   possibly it is a this pointer which is being somehow mismanaged in a higher-level function...

chuggafan:
I have the calling code here as well:
```assembly
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
add         eax,4
push        eax
call        00469656
pop         ecx
```
I think since the code in question is using a `__thiscall` calling convention eax+4 is correct... 
```assembly
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
push        eax
call        004685A6
pop         ecx
```
This is the code that calls *THAT*
and finally, the code that calls that code that calls the code that calls the issue:
```assembly
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
cmp         dword ptr [eax+4],0
je          00464B3D
mov         eax,dword ptr [ebp+8]
mov         eax,dword ptr [eax+4]
push        eax
call        004667B4
pop         ecx
pop         ebp
ret
```
Clearly there are two important issues at play:
1. There's something wrong here with the code and how it functions
2. Somehow this access with the reference translates into bad functions here.

The call stack causing this is listed again here for posterity, with the three relevant sections:
```
                        469662: std::__1::__libcpp_atomic_refcount_decrement<long>(long&) + 0xc  module: memory, line: 3386
                        4685b5: std::__1::__shared_count::__release_shared() + 0xf  module: memory, line: 3439
                        4667c0: std::__1::__shared_weak_count::__release_shared() + 0xc  module: memory, line: 3482
```
I currently am at a loss for why it would be crashing at that assembly as it seems that it is reasonable... maybe it's a triple indirection due to a `__thiscall` calling convention? I don't know

LADSoft:
the caller of that looks like this:

```
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
add         eax,4
push        eax
call        00469656
pop         ecx
```

this is trying to access the member __shared_owners to pass to the last function.   It is a member function, so [ebp+8] is the thisptr.   __shared_owners is the first and only member variable.   Normally it would be at offset 0 of the structure, but here we add 4 because the structure has a virtual function, and space had to be made to point to vtab and exception information.   So this snippet looks good too...

LADSoft:
the next one is

```
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
push        eax
call        004685A6
pop         ecx
```
this is from:

```
    _LIBCPP_INLINE_VISIBILITY
    void __release_shared() _NOEXCEPT {
      if (__shared_count::__release_shared())
        __release_weak();
    }
```

where __shared_count is the first and only base class of __shared_weak_count.    Since __shared_weak_count has no virtual functions the base class is loaded at offset 0, which seems to be what we have here.    So this one looks correct too...

LADSoft:
for the ~shared pointer we have:

```
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
cmp         dword ptr [eax+4],0
je          00464B3D
mov         eax,dword ptr [ebp+8]
mov         eax,dword ptr [eax+4]
push        eax
call        004667B4
pop         ecx
pop         ebp
ret
```

which corresponds to:

template<class _Tp>
shared_ptr<_Tp>::~shared_ptr()
{
    if (__cntrl_)
        __cntrl_->__release_shared();
}

here __cntrl_ is a pointer to __shared_weak_pointer which is the first member variable in shared_ptr;   again it would be at offset 0 in the structure.   The code seems to be using offset 0 as the thisptr for the call, so this code looks good too.

LADSoft:
it *looks* like the problem may just be we need to define_LIBCPP_HAS_BUILTIN_ATOMIC_SUPPORT somewhere.   I guess try that next and if we still have a problem I'll debug it...

by the way thank you for breaking all this down for me, it was a real time-saver on my end!

chuggafan:
> by the way thank you for breaking all this down for me, it was a real time-saver on my end!

No problem, I've been dealing with debugging this for the better part of 4? 5? yearish, which makes writing bug reports easier when you know the necessary information, and yhea, we do not define `_LIBCPP_HAS_BUILTIN_ATOMIC_SUPPORT` since that is assumed based on internal `<memory>` shenanigans instead of anything logical or reasonable like in `__config`

chuggafan:
Done doing a compile with a fixed version that uses atomics:
```assembly
push        ebp
mov         ebp,esp
mov         eax,dword ptr [ebp+8]
or          ecx,0FFFFFFFFh
lock xadd   dword ptr [eax],ecx
add         ecx,0FFFFFFFFh
mov         eax,ecx
pop         ebp
ret
```
is now the resultant code, which still has the *exact* same issue as before, just on the lock xadd instead... Somehow I feel like the reference being made here is causing an issue... All of the rest of the (relevant) code has the same disassembly here, so the problem might lie even further back into the assembly code if the code generated here is correct...

LADSoft:
I'll look at it tonite...   BTW you could look at the crash dump to see what EAX is at the time of the crash if you've a mind to lol...

chuggafan:
```
EAX: 004AD373  EBX: 0019F0A8  ECX: FFFFFFFF  EDX: 02540E45  flags: 00010286
EBP: 0019F090  ESI: 00000000  EDI: 00401000
 DS:     002B   ES:     002B   FS:     0053   GS:     002B
```

EAX at this time looks like a virtual pointer, I had forgotten that might be a necessary thing, it *looks* like it's trying to access in the map file:
`4ad322   std::__1::basic_string<char, char_traits<char>, allocator<char>>::__init(char const *, unsigned int)`
I don't know why it would be this value though... maybe up the chain we're getting an invalid offset into the structure?

LADSoft:
ok i made some progress; it looks like a misplaced destructor call is causing the problem.   I was looking through the assembly and the code for this line in maker .CPP seems to be the culprit:

```
; Line 282:         if (goalTime >= dependsTime && !RuleContainer::Instance()->OnList(goal, ".PHONY")) 
```

here when the first clause is satisfied it short circuits the second clause and doesn't do any of the logic in the body of the if statement;   but the destructor for the shared_ptr coming from RuleContainer::Instance() is called anyway.   Using whatever happens to be on the stack at the time...

Honestly I thought I had fixed this problem when it popped up earlier but I guess not...   

may take me a little bit to get it resolved...

LADSoft:
omake seems to work better with the checkin I just made...   it was a good exercise, destructors were not being handled at all properly in the presence of short-circuit logical operations...

chuggafan:
With current code as-is on my machine this fully compiles and successfully completes. I am going to now begin the testing for the futures code.

chuggafan:
Futures code fails fast, fails early, *and* fails hard. Gets STATUS_STACK_BUFFER_OVERRUN when trying to run it. I do not know why offhand this could be happening, and I am 99% sure that the futures tests also run into this, so I can still postpone this PR until #559 is complete.

LADSoft:
i'll take a detour and look at the threading tests then...   gotta do them sometime anyway :smile:

Meanwhile, I figured out one reason why programs sometimes hang when you compile them with OCC.   Apparently, if they crash before enough of the runtime is initialized to properly handle the exception that is a bad thing!

chuggafan:
Turns out I'm also having actual issues with the way I've implemented futures code, so I'm going to have to make some changes on that front (for some reason this std::shared_future is not cooperating)

LADSoft:
ok thanks for the heads up.   Maybe that causes some problems with the threading tests?   

I'm making progress, I was down to 72 threading tests failing last night.   Only 23 compiler errors, the rest were problems with actually running the tests.

I've been bogged down in a single issue today...   it is whackamole...  fix one thing and immediately something else comes to the front.    Think i know what to do to finish it but am wrapping up for today :smile:

LADSoft:
ok thanks for the heads up.   Maybe that causes some problems with the threading tests?   

I'm making progress, I was down to 72 threading tests failing last night.   Only 23 compiler errors, the rest were problems with actually running the tests.

I've been bogged down in a single issue today...   it is whackamole...  fix one thing and immediately something else comes to the front.    Think i know what to do to finish it but am wrapping up for today :smile:

chuggafan:
Well, I got that part fixed, but the issue now lies in me not properly handling errors, somehow there needs to be a "pass the chain up" action taking place to indicate errors and I'm not sure how to handle it properly...

LADSoft:
im not familiar enough with futures to have any advice...

chuggafan:
It's extremely fun because the way that I have in terms of debugging is: nothing at all due to the fact that it's throwing a system_error, which has no stacktraces.... So I have to instrument a ton of the system that I've changed in order to correctly test why this is happening.

*FUN*

chuggafan:
An update to what's going on in my land over here: I've been trying different ways to get to the bottom of this debug pile but trying to find out where a call to abort happened is near impossible with the way MSVC has instrumented its abort calls, especially because this is only happening on multithreaded systems, it may-or-may-not have to do with cleanup of the threading system I implemented and I'm moving in that direction to see how this goes...

LADSoft:
yeah multithreading is always a hassle.   I find it takes a lot of imagination to go debugging it!

tonite I got to the point where all the compile errors are gone from the thread tests.   Now to clean up miscellaneous asserts and crashes.    There are 70 of them...   i did see where one of the tests did an infinite recursion...   I'm hoping a lot of the issues are related :smile:

chuggafan:
Figured it out: something to do with shared_future and MSVC, very fun, figuring that out is my next step...

chuggafan:
Here we go, got it down to a simpler testcase after all other theories were done. All I'm relying on now are the threading systems to work (std::future, etc.)

chuggafan:
Tried to get this into a "usable" state in terms of being compatible with GNU make: no luck there, access violations are fun .-.

I did also realize that there's a big problem with the way that the threading system currently works in that we're spawning the thread before we actually ensure there are resources for the thread, so if there's more allocatable threads than cores (e.g. billions) we will literally crash the system, which is very fun, but I have no idea how to refactor everything to allow for that.

My personal preference for the refactor would be to use the JobServerAwareThread and have that handle the setup and teardown of its internal job server, but I don't know how to go about doing that, and it while it feels right, it also feels out of scope for this as it's been a long-running issue...

LADSoft:
oops, i was wondering what you meant but now I see the problem.   Yeah....

meanwhile I found out there is a special namespace 'this_thread'.    Probably some of the thread tests fail because it isn't handled correctly...    there is also something funky going on with the thread local data which I haven't sussed out yet...

chuggafan:
Quick question regarding MAKEFLAGS: I want to remove the number on -j for all dependant makes, as well as make it pass through the jobserver name via this, ideally this should actually be parsed from command line the job-server name and I just have a temporary one, but right now I'm doing things wrong, is there a way I can modify makeflags to actually be a bit useful here or will that take awhile?

GitMensch:
> I want to remove the number on -j for all dependant makes, as well as make it pass through the jobserver name via this, ideally this should actually be parsed from command line the job-server name

This is _exactly_ what happens with GNU make and what should be done - the first `make` will start the jobserver and pass it the number of parallel flags it uses, then removes that from `MAKEFLAGS` and start all sub-processes through the jobserver, so the sub-makes know about this and use the same jobserver - which task it is to control the number of parallel (actual running) make processes along with not starting any further process if `-k` was not given and one of the sub-makes fail.

I think that definitely would be a good approach for OrangeC make, too.

chuggafan:
> This is exactly what happens with GNU make and what should be done -

I wonder why I asked about it? ;)

It would still be nice to have the ability to pass in jobservers separately as well anyways because that way you can have a shared jobserver between different build systems.

LADSoft:
so there is a function SetMakeFlags which sets the MAKEFLAGS.    It creates a string and then assigns that to the MAKEFLAGS variable.  You can see it adds the -j: switch with a value a ways down.   

You could change it any way you want.   You might consider adding a new switch of type CmdSwitchString to hold the jobserver name and just stick it in the makeflags as well...   making the switch is as as simple as putting a prototype in the header then adding the switch where the other switches go.    The reference to the SwitchParser instance ties it in with all the other switches...    Then just use Get() to get the value...   there is also a GetExists() you can use to tell if the switch actually exists on the command line if you don't want to just go by whether or not the contents are empty...

LADSoft:
so I'm having a little problem with futures.    This code comes out of the tests:

```
        std::packaged_task<double(int, char)> p(A(5));
        std::future<double> f = p.get_future();
        std::thread(func, std::move(p)).detach();
        try
        {
            double i = f.get();
            ((void)i); // Prevent unused warning
            assert(false);
        }
        catch (const std::future_error& e)
        {
            assert(e.code() == make_error_code(std::future_errc::broken_promise));
        }
```
where the catch should be taken because something throws an exception.   OCC doesn't throw the exception though, it just hangs waiting for the promise to be fulfilled.    I looked a bit and it seems like it is the destructor ~promise() that throws the exception.    But I'm having a hard time seeing why that destructor should be called.   Any ideas on how this should work?

chuggafan:
The reason it's destructed is because it's moved to exist inside of the thread. The move operation destructs it on the main thread, and when inside of the thread A(5) does something(?), possibly never actually returning anything, causing the promise to be "broken", when searching it seems that there is a function there, but it's never called, so it doesn't initialize the promise with anything, thus, "breaking" the promise, and when the function exits, it is no longer owned, calling the destructor.

LADSoft:
ok that makes sense, i didn't catch that when future:get() exits it destructs the future.    But that puts me back to another problem, I'm not sure why it should exit?   if the future isn't 'ready' it goes waiting on a condition variable and that is what hangs.    So apparently for this to work the future has to be 'ready'.   But I can't for the life of me figure out what is supposed to make the future 'ready' so that get() would exit and call the destructor?

chuggafan:
The future is ready when the packaged_task's function gets called and returns a value into the promise, doing promise.set() or promise.set_at_thread_exit(). but because the package_task's function is never called, there is no value for get to return. get actually throws an error (in MSVC) if !valid, meaning that we need to have a valid return state, of which, if nothing is actually available yet, it isn't valid. 

my explanation's probably pretty bad but I'll be honest and say my thoughts originally were exactly the same as yours

chuggafan:
Apparently actually microsoft considers that test flaky due to reasons, [Listed here](https://github.com/microsoft/STL/blob/main/tests/libcxx/skipped_tests.txt) It's probably a really good idea to look at this for deciding what tests to actually work on...

LADSoft:
ill take that document into consideration if I am still having problems.  but meanwhile I did find a couple of bugs while I was playing with this...   the relevant one here being that functions should call the destructors for class instances that are passed in by value, and that isn't happening.   if it were the destructor for 'promise' would be called when the thread performing the packaged task exits, and possibly that would fix the issue.

i also found out that local statics (in functions) have to have thread-safe initialization since C++11.    So I'll be reworking how all that is done.

on the plus side there was a relatively simple bug in the 'exception' header that was possibly causing a lot of crashes in various tests.

LADSoft:
so fixing the destructor thing did fix the problem...   with the latest changes we are down to 55 failures on the thread tests....

chuggafan:
It looks like everything except the complaint I had about the issue with too many threads being spawned at once is all good now, and I don't know how to solve that issue, if Appveyor is happy with this, I'm happy with this, unless anyone can point out any issues.

I'm probably also going to pull out the infrastructure I created in order to make this possible into its own separate repository.

LADSoft:
other than the inno setup server failures, not sure what happened to the build.   I built it at home and it was ok...   anyway I can see a big difference in how omake responds now.   I kinda like it, it is a little more structured lol!

I will come back and see how things are tomorrow night...

chuggafan:
I also am not sure what happened, I'll try building it appveyor style sometime at another point to see if that has anything to do with it...

chuggafan:
Ah, I found an issue, I think it's very related to the issue I pointed out earlier, or my code is completely screwed with the changeover to MAKEFLAGS.... cuz 
hmmm
should never happen when I do -j:16, I'll investigate this at a later point than today as I have a lot of stuff to do today.

chuggafan:
I just realized the cause: MAKEFLAGS isn't actually setting any of the flags that OMAKE uses, so I was assuming wrong. I don't know enough of the infrastructure to fix something like that though...

LADSoft:
the problem is probably that the CmdSwitchXXXX variants don't set the 'exists' flag that GetExists uses...  so every invocation of omake is using its own local job server.  I'll fix it and do a push tomorrow night.

LADSoft:
i did make an attempt to fix this last night...

chuggafan:
Yhea, it worked, the main problem is now that due to using a regular mutex in some places the make program slowed down significantly, I'm planning on installing Intel VTune and profiling bottlenecks throughout the code to see how to fix this...

chuggafan:
Well, intel's VTune does help my postulation that we severely oversubsribe the CPU with threads, where:
```
Total Thread Count:	6,463
    Thread Oversubscription:	837.990s (58.0% of CPU Time)
Spin and Overhead Time:	504.085s (34.9% of CPU Time)
    Top Functions with Spin or Overhead Time
        Function	Module	Spin and Overhead Time	(% from CPU Time)
        SwitchToThread	KernelBase.dll	447.427s	31.0%
        WaitForSingleObjectEx	KernelBase.dll	24.912s	1.7%
        WaitForSingleObject	KernelBase.dll	21.921s	1.5%
        WaitForSingleObjectEx	KERNELBASE.dll	2.309s	0.2%
        WaitForSingleObjectEx	kernel32.dll	1.839s	0.1%
        [Others]	N/A*	5.677s	0.4%
```

This was calculated with a `make clean /j:16`, then with this information:
```
Collection and Platform Info
    Application Command Line:	C:\OrangeC\src\Release\omake.exe fullbuild /j:4
    Environment Variables:	
    Operating System:	Microsoft Windows 10
    Result Size:	578.5 MB
    Collection total time:	5:09
    Collector Type:	User-mode sampling and tracing
    CPU
        Name:	Intel(R) microarchitecture code named Alderlake-S
        Frequency:	3.6 GHz
        Logical CPU Count:	20
        Cache Allocation Technology
            Level 2 capability:	available
            Level 3 capability:	not detected
```

```
Effective CPU Utilization:	18.9% (3.780 out of 20 logical CPUs)
    Effective CPU Utilization Histogram
        Simultaneously Utilized Logical CPUs	Elapsed Time	Utilization threshold
        0	39.538049000099996	Idle
        1	52.256429499999996	Poor
        2	35.5800407	Poor
        3	19.3882666	Poor
        4	15.703936599999997	Poor
        5	14.640509899999994	Poor
        6	14.340884099999997	Poor
        7	13.524481699999999	Poor
        8	12.492332900000006	Poor
        9	9.446001800000001	Poor
        10	6.9649352	Poor
        11	5.059683300000001	Poor
        12	3.011359600000001	Poor
        13	2.3154806999999997	Poor
        14	1.8659705999999998	Poor
        15	1.367755	Poor
        16	0.9132054000000002	Ok
        17	0.3214995	Ok
        18	0.1352079	Ok
        19	0.04502679999999999	Ok
        20	0.0033439999999999998	Ideal
    Total Thread Count:	6,463
        Thread Oversubscription:	837.990s (58.0% of CPU Time)
    Wait Time with poor CPU Utilization:	24865.699s (99.2% of Wait Time)
        Top Waiting Objects
            Sync Object	Wait Time with poor CPU Utilization	(% from Object Wait Time)	Wait Count
            Semaphore 0xfe830145	7045.995s	99.3%	254
            Semaphore OMAKE28076 0x2662c13c	3190.251s	98.9%	207
            Thread	2522.803s	99.5%	1,781
            Process 0x61f1a82d	2440.384s	99.1%	1,119
            Thread 0x55d50b1f	1971.873s	98.6%	4,957
            [Others]	7694.394s	99.3%	294,591
    Spin and Overhead Time:	504.085s (34.9% of CPU Time)
        Top Functions with Spin or Overhead Time
            Function	Module	Spin and Overhead Time	(% from CPU Time)
            SwitchToThread	KernelBase.dll	447.427s	31.0%
            WaitForSingleObjectEx	KernelBase.dll	24.912s	1.7%
            WaitForSingleObject	KernelBase.dll	21.921s	1.5%
            WaitForSingleObjectEx	KERNELBASE.dll	2.309s	0.2%
            WaitForSingleObjectEx	kernel32.dll	1.839s	0.1%
            [Others]	N/A*	5.677s	0.4%
```

We wait *FAR* too much on semaphores while having threads to switch contexts too, it's actually disgusting. I suggest prioritizing fixing how we wait on semaphores in the context of this PR so that there is minimal issue created due to threading, there may also need to be considerations taken such as threads finishing up and calling dependencies, etc. And this *will* need to be fixed, if not now, for the future.

(PS: I get that it says we spent most of the time in 2 cores, but because that's due to me limiting the process to 4 threads instead of 16 that's more of what it should look like than what it actually will be compared to running full tilt on brand new top-of-the-line system such as mine).

LADSoft:
looks like a fix for the threading *might* be easy...   instead of doing the TakeJob in Spawner.cpp   introduce a TryTakeJob and put it in Runner::RunOne right before the if (touch) line...  if it can't take it should pop the rule stack and return -1.   That way threads won't be created if there isn't a job slot for them to actually run.    The release would be down in the last else after the join: label somewhere...

That is possibly much cleaner anyway as it will run all the commands for a rule before giving up the job slot....

If you want you can try that out...   otherwise I'll do it sometime this week and see how it performs...

chuggafan:
There's some issues right now where the entire OMAKE system seems to deadlock... I'm going to try to investigate why but this seems to be an incredibly complex multithreading topic....

LADSoft:
it is a little complicated...  I've been wondering if maybe it should be rewritten.    But basically, the way it works right now is that the function Maker::RunCommands is run, and it starts at the top of the dependency tree.   It calls Runner::RunOne for each dependency at the top of the tree.   Runner::RunOne never blocks, it either creates a thread or looks to see if a thread is done and then returns.   If it isn't doing much of anything it returns -1.    

In runner::RunOne the first if statement is taken if a dependency hasn't been started yet, which goes into a look checking all the sub-dependencies.   Again Runner::RunOne is called recursively on each subdependency.   When all the subdependencies are done it creates a thread for the current dependency, then the next invocation for RunOne doesn't take the first if but instead looks to see if any result is ready.   Until the result is ready it returns -1, then eventually it gets to the 'join' label and marks the dependency done and returns 0.

Again when all the subdependencies are done for a given dependency the first if completes and creates a thread for the current dependency.

The key to this is that RunOne never blocks but always returns either -1 which means still waiting, 0 which means the dependency completed normally, or a positive number, indicating that some command returned a non-zero value.   So it is run in a loop...   OS::Yield is used at the top level to give time to breathe.

What I was suggesting is that before the thread is created, take the job.   Release it when the thread completes.   That way you won't have a lot of outstanding threads.   

I did miss a key detail though, before we weren't taking the job if omake was being recalled recursively but with the change we are always taking the job even if there will be a call to omake...   and that would do things like deadlock (by having too many omake instances holding jobs) it or keep it from running a bunch of stuff simultaneously.   So your class TakeIfNotOmake should be left but the sense changed to GiveIfOmake to compensate...

LADSoft:
definitely needs a rewrite.   Probably uses a lot of CPU time spinning....   I think I know how to make it better but not sure if it would interfere with what you are doing.

meanwhile I prototyped the trywait thing... it is a little more complicated than i thought.   If you want I'll push the changes so you can see what I did...

chuggafan:
If you got a working prototype by all means, I made a lot of the base here so that it can be bolted ontop of the existing JobServer stuff...

LADSoft:
did the limiting of threads help anything?   OR is it still extremely slow?

chuggafan:
I didn't particularly test it since I was confused at the merge and I haven't had too much time, I'm still quite lost on why the Give and Takes are in certain areas... I'll probably get around to it sometime this week.

chuggafan:
Well, as it turns out it might just be the compilation process itself with OCC, as when I timed it last night with all of my CPU (20 threads) it took about a minute to do a fullbuild. This likely has to do with occ compiling occ being slow rather than anything else. So it's my bad for assuming a lot of things with this setup. I'll have to do more profiling though and see what VTune thinks for a bit but that will have to wait until I can wait 24 hours for VTune to get full results

LADSoft:
yeah i've tried a lot of things to speed it up and on an individual file basis it seems quicker than it was, for example a couple of the problematic files that took 8 or nine seconds to compile are now down to five.  But, regardless of what I do I can't make a dent in the build speed.    On my computer, running through one of the passes where it does three compiles followed by compiling the test hovers right at 15 minutes.

The one time I looked at timing it seemed like compiling a single file took significantly longer when omake was running than when it wasn't...     Although admittedly I didn't look at it very long.   But it does make me wonder what the effect is of all the spinning omake does.    And there are a lot of instances of omake running simultaneously so it is compounded.

I haven't looked into prototyping a fix for that as I've been busy with the containers tests....    and I'm also not sure what kind of collisions you and I would get into if I did that now rather than wait until you are comfortable with your changes and we can merge.   It seems like it might be a rather invasive change to rewrite the code that needs rewriting for this...

At one point I did try a build with mingw32-make just to see if there was any difference, but, after a couple of hours fighting with issues I went back to working on the tests without making any conclusions.

chuggafan:
I'm 99% sure that the reason the build is failing right now is a futures problem, but it's only appearing when OMAKE is compiled with OCC, so that might be the issue. *FUN TIMES* are about

chuggafan:
I FIGURED IT OUT! pulled out windbg and apparently an invalid name was being passed down, I have no idea how this is happening and am investigating it now.

chuggafan:
Ok, so this is actually a compiler bug in OCC where when performing a move operation the destruction of the type is happening before the move of the type, let's take the example here:
```
Creating a new job server
Creating job server with name: OMAKE4728
Closing handle for name: OMAKE4728
move assignment
Job server name NEW: OMAKE4728
Job server name: OMAKE4728
```
For context, the semaphore is instrumented with the "Creating job server with name:" in the constructor for creating a brand new server with a defined amount of threads, and the "Closing handle with name:" when performing a destructor. The move constructor call is also in with a custom move constructor I added.
With the move assignment happening *AFTER* the destruction of the job server, of course there's not going to be a valid handle for me to work with!
An investigation as to why this is happening should also be fixing a lot of other tests.

LADSoft:
if you haven't already, can you merge the mainline into your branch?   I've fixed some problems with constructors/destructors recently and want to verify whether the issue still exists.

chuggafan:
I already did, that's where the issue popped up for me, actually.

LADSoft:
i figured out what is wrong....   will try to fix it soon....

LADSoft:
the latest push seems to work a little better.   It got through an entire build for me.  It does seem a little slower when it gets to the point of compiling with the OCC compiled with itself....   but it seemed ok in earlier passes.    May be more of the same issue with the spinning compounded with the futures stuff you added, it is hard to say though since we don't know if the spinning is a big problem...

chuggafan:
Yhea, it looks like what you did was the trick, I'm going to do some investigation, but it just seems like -j:4 is extremely slow, I timed my compile just now and it took around ~12 minutes to perform, meaning that there's a significant amount of time that can be saved just by throwing far more powerful hardware at it if possible... it might be possible to just buy an old Dell rack and use that with windows 10 pro so there's docker containers, or if there's money to be thrown around playing around with windows server and a new threadripper/xeon type system.

But to be honest, I think our current situation is fine and trying to optimize the speed at which the build process runs is an ideal anyways, though we should figure out why the system takes *so long*, and stuff like github CI seems to have even *less* provisioned cores (2 cores), TeamCity seems like one that might work if you want to try, though you'd have to actually ask jetbrains for a license for that.

chuggafan:
On my recent fullbuild I had the exact number of semaphores get spawned that I wanted, so the only thing left to clean up now is the mingw access violation, which *WILL* be a pain.

LADSoft:
cool!   let me know if the access violation is another occ bug :smile:

chuggafan:
A quick reminder for me: OrangeC's compiles don't work inside of the /bin/bash replacement on windows for some reason... presumably due to the command never exiting? No clue why it does that, but that's something to take into account, and finding the batch files to run msys2 from cmd is annoying.

GitMensch:
> A quick reminder for me: OrangeC's compiles don't work inside of the /bin/bash replacement on windows for some reason... presumably due to the command never exiting? No clue why it does that, but that's something to take into account, and finding the batch files to run msys2 from cmd is annoying.

I guess I can help here, I just need to know what you actually wanted to say :-)
What do you mean with the compiles don't work? What with "inside of the /bin/bash replacement on windows"?
Presumably as an alternative, you want to call mys2 from cmd; or maybe just a single shell script / make invocation in this? I Is this about Appveyor? Is the cmd started from msys or "clean"?

chuggafan:
Ah, I meant it just calls into the command prompt when you attempt to do anything with the msys2 layer. So running make/omake/remake all fail because we're calling cmd.exe commands instead of /bin/sh commands, I *THINK*.

chuggafan:
So i've discovered: `make` with the `defterm` terminal doesn't work, but `omake` does, and `omake` and `make` don't work with `mintty` terminal, and I don't have conemu installed so I can't test it. *EXTENSIVELY* fun to debug something like that

chuggafan:
So in debugging: it's definitely my fault, the *REAL QUESTION* is how. Which will be a *LOT* of fun because GDB is not co-operating with me.

GitMensch:
> Which will be a LOT of fun because GDB is not co-operating with me.

I _may_ be able to help here if you specify how it does not co-operate. I found the MSYS2 gdb-multiarch x64 be quite helpful on Win32 (to debug everything), I mostly use that _outside_ of the terminal emulator and `attach` to the process (or use `target extended-remote` to connect to a gdbserver which runs the process I'm interested in).
Of course "normally", at least in breaking and attach scenarios one has to switch the thread (`info threads`, `thread N`).

GitMensch:
... and sometimes fall back to "WinDbg preview" (found in the MS store) to debug things GDB seems to be struggling with.

chuggafan:
WinDBG and GDB are both equally useless here, both are getting effectively the same thing:
`#0  0x0000000000000000 in ?? ()` as the backtrace

chuggafan:
Figured it out: it's futures, *AGAIN*. They *NEVER* seem to work, ever, 😢

GitMensch:
Do you see that bad backtrace in all threads?

chuggafan:
Only in the thread that was actually executing, the real issue is in gthreads, here:
```
__gthread_once (__func=<optimized out>, __once=0x1f4a321ac3c) at C:/msys64/mingw64/include/c++/11.2.0/x86_64-w64-mingw32/bits/gthr-default.h:700
700         return __gthrw_(pthread_once) (__once, __func);
```

GitMensch:
So the debuggers show everything nicely?

chuggafan:
Not when trying to get to the bottom of it easily, when the error occurs a SIGSEGV is sent and the stack pops all the way up in such a way that the stacktrace gets obliterated completely, leaving the stack trace looking like this:

```
Thread 5 received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
```

chuggafan:
It genuinely seems that it steps into pthread_once and immediately crashes. I have no idea how or why or how to fix this, extreme fun is about to ensue from code I didn't write and can't easily get a hold of to get a debug version of all of the symbols .-.

LADSoft:
FWIW, I find that usually if a stacktrace has an EIP of zero, it means there has been a call through a null function pointer...   I've specifically had those kinds of problems where bugs in OCC would cause call_once and friends to do that sort of thing...    it can be a real hassle to debug it.   But it seems like the issue here isn't occ...

I notice from the stack trace you were able to get this is when a new thread is being started, probably using thread() since tuples are involved, perhaps implicitly by futures.    maybe somewhere in the threading/futures you aren't initializing something quite like the standard says to, or are otherwise invoking 'undefined' behavior, and it is causing issues with mingw64's implementation of the standard?

chuggafan:
```
rax            0xd                 13
rbx            0x1da79faac3c       2037860969532
rcx            0xffffffff          4294967295
rdx            0x0                 0
rsi            0x22bf9ffd40        149243821376
rdi            0x1da79fa6190       2037860950416
rbp            0x22bf9ffd60        0x22bf9ffd60
rsp            0x22bf9ffbf0        0x22bf9ffbf0
r8             0x7ff90e185940      140707660060992
r9             0x22bf9fdf00        149243813632
r10            0x0                 0
r11            0x246               582
r12            0x1da79faad10       2037860969744
r13            0x0                 0
r14            0x0                 0
r15            0x0                 0
rip            0x7ff67fc6269e      0x7ff67fc6269e <std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)+94>
eflags         0x206               [ PF IF ]
cs             0x33                51
ss             0x2b                43
ds             0x2b                43
es             0x2b                43
fs             0x53                83
gs             0x2b                43
```

And the function pointer is: 7FF67FC52720 so it definitely seems valid... I'm not sure if there's undefined behavior in there now but if futures are this hard to actually use why were they ever included in the standard? Seriously...

LADSoft:
I looked briefly at this - as a first attempt I built everything with mingw64 then copied the resulting omake.exe over.    Doesn't seem to crash but definitely fails - when built from scratch with any more than one thread selected.   -j:1 works fine...   -j:2 doesn't.   Can't resolve the targets...

I also found that if you do a build of the MS version with -j:1 and then delete some of the resulting executables so that another build would have something to do, -j:8 works fine and rebuilds the executables.

So that leads me to believe this is likely to be a timing problem, exacerbated by all the instances of omake that are running when you build from scratch with an omake built for the x64 platform....    perhaps some access is not fully guarded against multiple threads.   It is possible that if you enable the 64-bit build in visual studio you will see a similar problem in the debugger...   but you never know about stuff like that when timing problems are involved.

If you want I'll dig into it further over the weekend...

chuggafan:
Please do, I'm short on time this weekend sadly so not much debugging work going on there. If something is not fully gaurded against multiple threads I'd actually be legitimately surprised, yet not surprised, and those types of bugs are *PAINS* to debug.

chuggafan:
Did you get anywhere with this or should I try to file a bug report with the mingw people and see if they're the ones having issues since I can't find the winpthreads source code to make a debug library with it.

chuggafan:
Turns out you're right, it crashes out after calling the function pointer for some reason, I need to investigate why the function pointer isn't working and see if this is a glibc issue or if it's an us issue...

Finally figured out how to debug this too (I needed to actually set a breakpoint in the file I was referencing to get it to work)

chuggafan:
I found the exact address in libstdc++6.dll that it happens, but because there's no debug information, I can't easily debug this far.... *EXTENSIVELY* fun.

`0x00007ffcd4a40fc7 in ?? () from C:\msys64\mingw64\bin\libstdc++-6.dll`

GitMensch:
I've rechecked with the MSYS2 team. At least all of the core components are compiled without debug information.
The only way to adjust this is to fork MSYS2 and trigger a build of gcc with an adjusted package like https://github.com/jeremyd2019/MINGW-packages/commit/77519f06eddf5afe5ab1899195fc6e29cadbede1 - if I understood this correctly this should normally trigger the CI build enabling you to download and replace the files in your local MSYS2 environment.

Otherwise download the package definitions, do the adjustments above and just build locally.

chuggafan:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf

According to C++14 for the set_value function:
```
 Effects: atomically stores the value r in the shared state and makes that state ready (30.6.4).
Throws:
— future_error if its shared state already has a stored value or exception, or
— for the first version, any exception thrown by the constructor selected to copy an object of R, or
— for the second version, any exception thrown by the constructor selected to move an object of R.
Error conditions:
— promise_already_satisfied if its shared state already has a stored value or exception.
— no_state if *this has no shared state
```

It seems like I'm not triggering undefined behavior here. 

As best as what I can tell is going on is that it seems that gcc is optimizing out the `__Prepare_execution` class in their mutex implementation:

```cpp
  /// Invoke a callable and synchronize with other calls using the same flag
  template<typename _Callable, typename... _Args>
    void
    call_once(once_flag& __once, _Callable&& __f, _Args&&... __args)
    {
      // Closure type that runs the function
      auto __callable = [&] {
	  std::__invoke(std::forward<_Callable>(__f),
			std::forward<_Args>(__args)...);
      };

      once_flag::_Prepare_execution __exec(__callable);

      // XXX pthread_once does not reset the flag if an exception is thrown.
      if (int __e = __gthread_once(&__once._M_once, &__once_proxy))
	__throw_system_error(__e);
    }
```

Specifically the once_flag::_Prepare_execution section.

Because this is being optimized out, this is causing the issue, so the issue seems to lay at GCC's hands here, as for a fix other than "Don't use std::future", idk, and idk the specifics of what are causing this other than it seems that this is the root cause of the issue here because of the many tests I've performed. I've waited this long to confirm through three different GCC 11 versions to see if this is an issue and it is. I have to do a thorough analysis sometime soon but all signs point to GCC being at fault and not me here.

LADSoft:
given this is GCC's issue, can you think of a workaround we could make in our code that would allow us to bypass the issue?   I'd like to incorporate these changes but am hoping we don't have to get rid of the GCC builds to do it...

GitMensch:
personal vote: if known to be broken on GCC:  use an alternate version (and if it is slower #ifdef that)

chuggafan:
I don't think it's possible to go around this, and I'm only 99% sure this is GCC's issue since we don't actually run on Linux, the correct place to report this would be to mingw so they could report to to GCC as it could also be a MINGW issue. I can just disable this version for now and we can use a switch to swap back to the known working version of this as much as I might dislike it for now and simply keep it there until the platforms properly support this.

LADSoft:
i'm not 100% clear on what you want to do....    are you suggesting keep this PR open until it gets fixed?     Or do you want to merge this PR and just drop the mingw64 build until it gets fixed and appveyor has a new version of mingw64?    

I guess another possibility is not have the mingw64 build use version of omake it builds so we don't run into the issue, although that would mean reworking the build script?

chuggafan:
I'll probably just leave the newer codebase in for now and let the old codebase run with some defines, newer GCC versions also have this issue and I'm not currently sure how to make a small but reproducible build of this issue for GCC. I know there's a tool but I don't remember its name anymore.

chuggafan:
I am genuinely shocked btw that the assembly compile failed, I have no idea and the ASSEMBLYcompiles.bat stuff didn't work to replicate it so I have to go through and build exactly as appveyor does it. I'll get around to that sometime this week.

chuggafan:
Still haven't been able to replicate the failure on my side, not exactly sure what the heck went wrong here....

LADSoft:
since occopt crashed,  I've started a new build to see if it is going to happen again...

chuggafan:
I think you rebuilt the wrong PR....

chuggafan:
I have actually no idea why the CLANG WITH STATIC LIB target suddenly gained an extra 7 minutes for the compile, probably due to extra pressure on the appveyor servers.

LADSoft:
@chuggafan thank you for all your work on this PR!

LADSoft:
going to close this issue as omake has been working for several months.    I will make a separate issue for getpid and getppid...

PR #667, opened 2022-01-15
minor editorial for debughelper

GitMensch:
* add note about `%PATH%` (which is important for running generated programs outside of the compiler)
* format all switches identical
* fixed a typo of `aa1()`

feel free to close and do other/minor adjustments to the file as you see fit


LADSoft:
thank you for your edits!

PR #678, opened 2022-07-15
Fix rare issue where oasm could be potentially generating incorrect code

chuggafan:
I noticed this issue where xleft->FVAL + xright->FVAL were being pointed out to me by Visual Studio during an Analysis run, as it turns out the issue at hand is that the FVAL property can be accessed due to an enum being a part of the class. I checked to make sure no other instances of something like this was happening, and thankfully this is the only case.


LADSoft:
I was just going to rebuild #660 one more time before calling the issue you had a fluke and found you doing this...   possibly there is some inter-relationship between this and what was going on there?

chuggafan:
I doubt that, I think it could be just a fluke, not entirely sure, it happened pretty soon into that occopt call too.

This is just an oasm issue AFAICT which could affect things, but I'm surprised we hadn't found something like this sooner.

PR #679, opened 2022-07-22
Add fixes to improve many issues with current libcxx tests

chuggafan:
The fixes here are for a lot of tests failing not having correct support
in creating a temporary file name, tests for inttypes not having all of
the correct integer formats for C++, and basetypes not being checked
when searching for candidates' symbol pointers, causing crashes when
types are const or volatile.

I'm also working on an issue where noexcept specifiers on specific constructors aren't being applied correctly and am trying to track down in-code where that can be happening but due to the sheer number of places debugging that will probably take a few days.

### DIFF
--- a/src/clibs/cpp/libcxx/test/support/platform_support.h
+++ b/src/clibs/cpp/libcxx/test/support/platform_support.h
@@ -58,6 +58,7 @@
 #include <string>
 #if defined(_WIN32) || defined(__MINGW32__)
 #include <io.h> // _mktemp_s
+#include <fileapi.h>
 #else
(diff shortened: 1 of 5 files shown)


GitMensch:
Just to make sure I understand it: as this is not a draft it is thought of as "directly merge in", correct? (Looks good to me)

chuggafan:
Yep, the other issue I mentioned will take a lot longer to solve than these and isn't directly related so I threw this up for now, I'm also working on fixing the stuff I mentioned inside of the recent omake commits about linux semaphores.

Issue #701, opened 2022-09-10
fix release page and drop superflous tags


GitMensch:
There are four tags for v6.0.61.1 as seen in https://github.com/LADSoft/OrangeC/tags (all pointing to the correct hash [09885caa4522dad04ab60207149d3970e104130f]) - please drop three of them and force-push (ideally also doing for the older releases v6.0.45.1 and v6.0.44.2 which have two tags).

This will also cleanup https://github.com/LADSoft/OrangeC/releases which currently has 3 entries for the current release.

To close the issue the only thing left then would be to add the release note entry of the current version from https://github.com/LADSoft/OrangeC/blob/master/HISTORY.md (oops, also missing there).

GitMensch:
What a great release!

PR #702, opened 2022-09-10
markdown cleanup HISTORY

GitMensch:
This is no complete work, but fixes the most obvious issues, it now renders "well enough" and has appropriate sub-headings which can be linked, to.

I hope this may help a bit with the missing entry for the current version spotted with #701 (of course, also in src/releasenotes.txt)

One thing that is open:
Os it Frank or Franck? What were his referenced changes?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,6 @@
-Release notes:
+# OrangeC Compiler And Tool Chain Release Notes
 
-version 6.0.60:
+## Version 6.0.60:
 * occ: various template handling fixes
 * occ: various constexpr fixes
 * occ: various noexcept fixes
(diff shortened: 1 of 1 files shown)


LADSoft:
I think it is Franck.    He didn't really make changes but had a lot of comments about bug fixes and how to generate better code.   He was running lots of code through the compiler and found various bugs.    This is long before I made it to GitHub.   He is the one that got me pulling down other peoples code and running it through occ...    the two main things I recall is he coached me into how to generate floating point compares that would be sensitive to NAN's, and that is where the 'proper' handling of include file lookups came from.    I think he also helped me bring the preprocessor up to standard...    he may also have helped work out issues with implicit conversions...    but there was lots of stuff he found...

GitMensch:
> I think it is Franck. 

Would you mind adjusting HISTORY.md and src/releasenotes.txt on this piece? Possibly the file wanted to say in the second (=older) reference that this version's changes effectively are based on his suggestions - then just drop the leading `* ` and possibly adjust the text a bit.

PR #803, opened 2023-04-02
Add beginning stages of C2x support

chuggafan:
[[__attributes__]] and [[attributes]] support mainly along with elifdef and elifndef We still have a long way to go but this basic support structure should be there

Also there's a bug with compiling this fixes along the way. Also, nullptr, can't forget that thing
And the fact that you don't need every keyword type We'll probably have to slightly modify the internal workings based on what the actual C standard deprecates (such as _Noreturn) later
This also fixes a compilation error related to the renaming of wininc

### DIFF
--- a/src/clibs/platform/win32/lscrtl/makefile
+++ b/src/clibs/platform/win32/lscrtl/makefile
@@ -64,7 +64,8 @@ DLLOBJS = $(shell dir $(B) $(CLIBOBJECT)\\*.o)
 	$(CC) /c /C-E $(CFLAGS) $(STDINCLUDE) -o$@ $^
 
 $(RCFILE).res: $(RCFILE).rc ..\..\..\..\version.h
-	orc -i..\wininc $(RCFILE).rc
+	echo $(CURDIR)
+	orc -i..\inc $(RCFILE).rc
 
 #DISTRIBUTE:
 #	$(ZIP) -porangec a $(DISTADDON)\lscrtl $(LSCRTLOBJECT)\*.*
--- a/src/occopt/beinterfdefs.h
+++ b/src/occopt/beinterfdefs.h
@@ -58,6 +58,7 @@ typedef struct
     bool prm_cmangle;            /* add underscore to names */
     bool prm_c99;                /* C99 mode */
     bool prm_c1x;                /* C1x mode */
+    bool prm_c2x;                /* C2x mode */
     bool prm_cplusplus;          /* C++ mode */
     bool prm_xcept;              /* generate RTTI in C++ mode */
     bool prm_icdfile;            /* dump intermediate code peep list to file */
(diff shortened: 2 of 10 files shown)


LADSoft:
the main thing I noticed is you added flags to '_Bool' to deprecate it in C2x, but prm_c1x is being set as well as prm_c2x in C2x mode, so, _Bool might not actually be deprecated the way you want...   also if you want to make an enumeration out of all those c version flags you are welcome to do it...

chuggafan:
While `_Bool` is declare obsolescent with the advent of C2x, I am not currently attempting to add warnings, etc. for marked obsolescent items, especially as this is really just a "first pass" and a more appropriate method of marking these constructs as obsolescent must first be devised before continuing on with the process of marking these constructs as warnings, perhaps even with source code transformations to ensure C99 compliance by including stdbool where appropriate and renaming the item...

The C version flag enumeration does sound interesting to do at some point, but I was not focused on that so much here...

LADSoft:
can I ask for a change?   I think we need to be a little careful about always declaring 'bool' as a type in C language programs, because, I suspect that there are a lot of legacy programs that either typedef it or #define it...    declaring it as a type in the C2x mode is great but in C99 for example it could be problematic.

The CC386 test in tests\occil\cc386 uses it as a variable name, which is why that build failed...

chuggafan:
I specifically made it so that it was a keyword only when it is used in C2x, so unless there is a switch of `/2` being used for the CC386 tests it shouldn't be an issue

`    {"bool", 4, kw_bool, KW_CPLUSPLUS | KW_C2X, TT_BASETYPE | TT_BOOL},` here is the line that I wrote, so that shouldn't be the issue, I'll investigate further however....

chuggafan:
I'm not sure we're seeing the same thing, I can compile:
```c
#include <stdlib.h>
#include <stdio.h>

int main()
{
    #define true 1
    _Bool bool = true;
    printf("%d\n", bool);
}
```
just fine, with:
```
C:\OrangeC>occ /1 test.c
occ (OrangeC) Version 6.0.62.2
Copyright (C) LADSoft 2006-2023
Warning( 75)  test.c(7):  'bool' is a C99 or C11 keyword - did you mean to specify C99 or C11 mode?
Warning( 75)  test.c(8):  'bool' is a C99 or C11 keyword - did you mean to specify C99 or C11 mode?

C:\OrangeC>test.exe
1
```

LADSoft:
no that is not what I meant although possibly it shouldn't be generating those warnings either if it didn't know bool as a keyword...

I was thinking more along the lines of:

```c++
typedef int bool;
```
or

```c++
int bb(int bool)
{
}
```

both of which complain about two many types in declaration no matter which flavor of the compiler you use, but bool shouldn't be a builtin type until c2x so both of those should succeed except in c2x or c++...   

I'm concerned about this because I can see where it might be rather common for people to have done the first of those two fairly often in older programs...

LADSoft:
your example doesn't compile for me either though same problem...   i'm running off the 'master' branch after the merge...

chuggafan:
Maybe a change I made but didn't push caused the problem? I am confused if you're getting a different error from me...

PR #804, opened 2023-04-14
Fix the C2X issue

chuggafan:
Apparently the error was being generated from the count not being high enough, when I merge this with the mainline it works perfectly.
This also contains a fix for a premature move with MSVC.

This also adds the `'` separator to C2x


LADSoft:
thanks.   the occil build passed so I guess it is fixed...   There was something up with the MSVC/static library build so I'm letting it build again to see if that is repeatable...

PR #857, opened 2023-07-08
Fix semaphores by just ignoring them on linux

chuggafan:
I've noticed you're playing around with semaphores.h, if there's any issues with it compiling on linux it's *not* intended for linux anymore, the main issue is fundamentally that linux semaphores don't function the way we need them to, thus the POSIXJobServer which is pipe-arbitrated

### DIFF
--- a/src/omake/JobServer.cpp
+++ b/src/omake/JobServer.cpp
@@ -31,10 +31,12 @@ std::shared_ptr<JobServer> JobServer::GetJobServer(const std::string& auth_strin
     return std::make_shared<POSIXJobServer>(readfd, writefd);
 #endif
 }
+#ifdef _WIN32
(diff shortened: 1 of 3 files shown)


LADSoft:
yeah the only reason I was playing with it was to get it to compile, since I was doing a linux build.   I have no intention to make it actually work at this point.    I did eventually get it so you can withdraw this pull request.

chuggafan:
Understood, I just noticed it and wanted to help by taking the easy route.

chuggafan:
It's only used on Win32 platforms, theoretically we keep it around so that it's easy to develop on both however... What I should do is make it a header/implementation style so that it's just stubs on linux but I'm a bit lazier than normal at the moment.

Issue #907, opened 2023-07-13
Support building Orange C with MinGW-w64 based Clang


iahunq:
`clang.mak` is for MSVC based Clang. We need to add a `llvm-mingw.mak` makefile for MinGW-w64 based Clang. The content of `llvm-mingw.mak` could be adapted from `mingw64.mak`.

You could download a MinGW-w64 based Clang from https://github.com/mstorsjo/llvm-mingw/releases (remember to pick the MSVCRT build).

Related issue: #728

Update: it seems we don't need a separate makefile at all, for building with MinGW-w64 based Clang, simply edit `mingw64.mak` and replace `-lstdc++` with `-lc++` on `COMPLIB`.

GitMensch:
Why would we need a separate one?

The MSYS2 packaging system uses a single definition for all its environments and those use the same makefile, they just pass CC and others as variable into make (which has gcc as most reasonable default, which on the ucrt and similar environments is just a link to the clang compiler).

iahunq:
> Why would we need a separate one? The MSYS2 packaging system uses a single definition for all its environments and those use the same makefile, they just pass CC and others as variable into make (which has gcc as most reasonable default, which on the ucrt and similar environments is just a link to the clang compiler).

Clang will use `libc++` as default even though you could use `-stdlib=libstdc++` to force it to link with `libstdc++` instead. On `mingw64.mak` you could see `-lstdc++` as part of `COMPLIB`. On llvm-mingw distribution or MSYS2 CLANG64, `libstdc++` is not even included.

iahunq:
Simply editing `mingw64.mak` to replace `-lstdc++` with `lc++` doesn't work. So, you can't use `mingw64.mak` as is.

iahunq:
@LADSoft It seems I have wrongly blamed the makefile. Please have a look:

```
In file included from memory.cpp:28:
..\occopt/memory.h:122:31: error: use of undeclared identifier 'beLocalAlloc'
    auto rv = static_cast<T*>(beLocalAlloc(sz * sizeof(T)));
                              ^
..\occopt/memory.h:137:31: error: use of undeclared identifier 'beGlobalAlloc'
    auto rv = static_cast<T*>(beGlobalAlloc(sz * sizeof(T)));
                              ^
```

I have no idea why the same piece of code compiled fine on GCC.

iahunq:
[...]

PR #915, opened 2023-07-13
Fix building with MinGW-w64 based Clang

iahunq:
It turned out MinGW-w64 based Clang needs the same treatment as the Embarcadero Compiler.

### DIFF
--- a/src/occopt/memory.h
+++ b/src/occopt/memory.h
@@ -24,7 +24,7 @@
  */
 
 #include "ctypes.h"
-#ifdef BORLAND
+#if defined(BORLAND) || (defined(__MINGW32__) && defined(__clang__))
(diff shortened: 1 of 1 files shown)


src/occopt/memory.h
```diff
@@ -24,7 +24,7 @@
  */
 
 #include "ctypes.h"
-#ifdef BORLAND
+#if defined(BORLAND) || (defined(__MINGW32__) && defined(__clang__))
```
GitMensch, line 27: Hm, what actually is buggy there and where should those functions be defined?

iahunq: @GitMensch See: https://github.com/LADSoft/OrangeC/issues/907#issuecomment-1634641363

PR #918, opened 2023-07-14
Add strnlen

### LINKED ISSUE #913: FR: implement strnlen
> That's a POSIX extension, not standard C, we can add support for it, but I haven't seen it before.

_Originally posted by @chuggafan in https://github.com/LADSoft/OrangeC/issues/278#issuecomment-1633385255_

chuggafan:
Super simple implementation, I used memchr instead of rolling a loop normally because memchr seems like it might be slightly faster (we should honestly test our codegen vs our assembly for our functions at some point though...) Fixes #913


LADSoft:
i had already added strlen and wcslen by the time I saw this; it will be in the next build.    Basically I cloned the assembly language code for strlen and put in max_length checks...

PR #1021, opened 2024-02-06
removed OCIDE from installer

### LINKED ISSUE #1020: IDE program not properly installed in new install of v67.3.1 on Windows 10 as current user
I installed a new install as a user install on Windows 10 x64 22H2 from `WindowsInstaller6738.exe`.

It seems the install did not correctly install `ocide.exe`
When run the IDE was checked on the installer, it showed this error.
install 6738 windows 10 error

the start shortcuts are dead, because `ocide.exe` is missing from `bin/` directory.

**Workaround:** manually extract the prog from https://github.com/LADSoft/OCIDE

GitMensch:
fixes #1020

additionally moved the examples to its own installer option, installing it directly below OrangeC

warning: untested...


LADSoft:
@GitMensch

there is a problem with these commits: you've changed various CPJ files to refer to where the files exist in the source tree, however, the files don't get installed there by occ.iss.   But that speaks to a broader issue because the files are installed in one place for desktop apps and in another place for thumbdrive apps, but, the CPJ files have fixed paths in them.    Possibly, the CPJ files need relative paths (from the location of the CPJ file) to take care of all three use cases,  which I think is something we added to later versions of OCIDE but don't remember for sure...

LADSoft:
im doing the gitignore and dosxclude.lst separately in another commit, I'm going to close this one

GitMensch:
e340548d8247a24b2b0b615045c4b90bc06c3593 handled the two files - but the removal of the OCIDE parts and the adjustment for the installer concerning the examples are still missing.

LADSoft:
so will fix this  and the examples thing in conjunction with closing #1020...

PR #1022, opened 2024-02-08
Update occil.md

GitMensch:
textual and formatting changes

... I'm definitely confused about CSC...


LADSoft:
csc is the csharp compiler, needless to say it must be on the path before you run the tests lol...    I think it comes with windows, but, I usually get it by running the Visual Studio command line script...

Issue #1031, opened 2024-05-08
Utils::AddExt not compatible with source files with more than one dot


rochus-keller:
My reference project includes a lot of files which include more than one dot, e.g. som.Vector.aa9eee315e.c
Here is the project: [awfy.zip](https://github.com/LADSoft/OrangeC/files/15250997/awfy.zip)

If I compile som.Vector.aa9eee315e.c with my orangefront and the -Y option, I just get one output file called som.Vector.aa9eee315e, which apparently is the icd version. The icf version is likely called the same and thus overwritten.

I traced the issue to Utils::AddExt, which does nothing in case of filenames like som.Vector.aa9eee315e.c; Utils::StripExt has worked correctly and left som.Vector.aa9eee315e, but then Utils::AddExt concludes that there is already an extension and does nothing.

I tried with this implementation instead:

```
void Utils::AddExt(char* buffer, const char* ext)
{
    char* pos = (char*)strrchr(buffer, '.');
    if (!pos || strcmp(pos,ext) != 0 )
        strcat(buffer, ext);
}
```
With this modification allone it doesn't work (instead it complains that it doesn't find file.c.C). To make it work, I also had to remove `Utils::AddExt(buffer, ".C");` from line 567 in occparse.cpp. 

With this modification, when I run with a single .c file, the .icf and .icd are generated correctly. But when I run with *.c, then all .icd files are generated, but an .icf is generated only for the first file (in my case Benchmark.icf). No indication of a problem from command line outputs. 

So apparently I didn't get the logic yet. Any ideas?

rochus-keller:
Meanwhile I found the reason why only one .icf is written if the command line says *.c. The problem seems to have been unrelated. 

To fix it, I had to move and modify the if clause from line 812 of occparse.cpp to line 714, see https://github.com/rochus-keller/OrangeC/commit/45bb9032834cac2f7e1ceefff45d71a0ddccc817.

But there is still the question whether I considered all effects of the change to Utils::AddExt. There is e.g. the line 1021 in osutil.cpp which still states `Utils::AddExt(buffer, ".c");` and which maybe has to be removed. 

What do you think?

LADSoft:
so i think occ does handle this somehow but it probably somewhere deals with it by looking for "..\\"  which is not what you want

I wasn't able to immediately see where it is taken care of but tomorrow I will look at it and figure it out.

[...]

PR #1032, opened 2024-05-09
Support base file names with dots and icf output for all provided source files (not only one)

rochus-keller:
This is the continuation of https://github.com/LADSoft/OrangeC/issues/1031 which is hereby closed.

I checked where the functions using Utils::AddExt are used and came to the conclusion that my modification of Utils::AddExt works up to the few cases where a default suffix should be added when the user provides a filename without suffix. But from my humble point of view this case should rather be treated as an error; it might have been a decent behaviour as long as Orange only was a C compiler, but today a plethora of possible suffices is expected, so it doesn't make sense to just arbitrarily select one. 

Conclusion: the compiler doesn't support filenames provided by the user without suffix anymore, and doesn't change the suffix of a user provided filename (as also suggested by [GitMensch](https://github.com/GitMensch)). And the generation of output filenames by chopping the suffix and adding a new one is supported.

I also modified the code order in occparse so the icf output is in the loop through all provided source files (the original version instead wrote the icf of the last file with the name of the first file).

### DIFF
--- a/BUSY
+++ b/BUSY
@@ -68,6 +68,9 @@ let common : SourceSet {
         ./src/util/Utils.h
         ./src/util/xml.h
 
+        ./src/ocpp/Floating.cpp
+        ./src/ocpp/Floating.h
+
         ./src/occopt/config.cpp
         ./src/occopt/config.h
         ./src/occopt/configx86.cpp
@@ -143,7 +146,6 @@ let orangefront : Executable {
         ./src/occparse/unmangle.cpp
         ./src/occparse/wseh.cpp
         ./src/ocpp/Errors.cpp
-        ./src/ocpp/Floating.cpp
         ./src/ocpp/InputFile.cpp
         ./src/ocpp/MakeStubs.cpp
         ./src/ocpp/PipeArbitrator.cpp
@@ -213,7 +215,6 @@ let orangefront : Executable {
         ./src/occparse/winmode.h
         ./src/occparse/wseh.h
         ./src/ocpp/Errors.h
-        ./src/ocpp/Floating.h
         ./src/ocpp/InputFile.h
         ./src/ocpp/MakeStubs.h
         ./src/ocpp/PipeArbitrator.h
@@ -282,16 +283,11 @@ let orangeopt : Executable {
         ./src/occopt/istren.h
         ./src/occopt/localprotect.cpp
         ./src/occopt/localprotect.h
-        ./src/occopt/msilprocess.cpp
         ./src/occopt/optmain.cpp
         ./src/occopt/optmain.h
         ./src/occopt/optmodulerun.cpp
-        ./src/occopt/rewritemsil.cpp
-        ./src/occopt/rewritemsil.h
         ./src/occopt/rewritex86.cpp
         ./src/occopt/rewritex86.h
-
-        ./src/ocpp/Floating.cpp
      ]
 }
 
--- a/BUSY
+++ b/BUSY
@@ -38,7 +38,7 @@ let config : Config {
 
 let config2 : Config {
     .include_dirs += [ . ./src/occopt ./src/util ./src/ocpp ]
-    .defines += [ "ORANGE_NO_MSIL" "ORANGE_NO_INASM" ]
+    .defines += [ "ORANGE_NO_MSIL" "ORANGE_NO_INASM" "ORANGE_NAMES_WITH_DOTS" ]
 }
 
 let common : SourceSet {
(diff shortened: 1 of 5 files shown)


LADSoft:
thanks for looking at this!

as some history, i was originally copying the old borland compiler....   back in the day when C++ was first coming out it still appended the .c automatically.   I never revisited that behavior.

It looks today  that most major compilers pass files without extensions on to the linker....   which then can make some assumption on its own about the nature of the file.    on windows CLANG and MSVC assume it needs a .obj extension, gcc thinks it is a linker script.    

I'm ok with just bailing for such files instead of making assumptions about them, even thought it could conceivably break make files,  but there really ought to be some error if we do it...   we shouldn't just be blithely ignoring it.    If that error comes from trying to compile it directly as specified that is fine I guess....   we just need something to say that this file couldn't be processed as a source file...

rochus-keller:
Ok, I see. As far as I understand the code in occparse main, a file with no extension would just implicitly be treated as a C file; without my modifications, a file without extension was just given the extension ".C", but there is no check on this extension anywhere, so apparently a C file is assumed by default (with or without the .C extension). I will check in the debugger what actually happens with my modifications if a file has no extension.

rochus-keller:
Meanwhile I conducted tests on Windows 11 and Debian 12 x86 with the most recent commit posted here. 

The frontend behaves as assumed. 

If I pass a file without extension, it is assumed to be C and compiled as such. The generated icd and icf look decently.

If I pass the filename without extension, but the actual file has a .c extension, the frontend complains that it cannot find the file with the passed filename. And vice versa (passing with extension but the file on disk has none).

So from my point of view the modifications already meet your requirements stated above.

LADSoft:
thank you for testing it!

that is fine, i don't mind it not being able to compile it i just want to know it didn't compile it lol....   I will accept these changes later on :)

rochus-keller:
Changing code that I don't know and where I can't estimate the full effect of the change is always a challenge, especially since I only have a subset of the code in my development environment.

I have a theory as to what the problem might be and have uploaded a new commit. Actually there was a section in osutil outputfile the purpose of which I didn't understand (and placed a TODO in the comment), but now I think that's the place where it fails. I can imagine that in the build the output file is specified without extension, and therefore it does not work because I had commented out the part. But that's just a guess, as I can't test it completely on my end.

LADSoft:
lol i guess i understand your frustration, there are a lot of unknown dependencies in this source code...

since i really want the dots thing to be fixed, could you revert your changes from this morning (you can use a force push if you need to) and I will figure out what the problem is and fix it?

rochus-keller:
No frustration, just stubbornness ;-)

Unfortunately even now when I revert to the original suffix logic ([329342d](https://github.com/LADSoft/OrangeC/pull/1032/commits/329342d94e3d9a849eab252fd0a80553b939ae4c)) the windows build fails. It started to fail with [abb3393](https://github.com/LADSoft/OrangeC/pull/1032/commits/abb3393a10b801e2087d49df1eec9e5724501a95), but since I reverted this (besides the .c instead of .C), I would have assumed that it should work.

Unfortunately, I'm not such a great git wizard, so I'd better leave it with the history change before I break something.

LADSoft:
ok lol yeah first time i did a force i had to be told exactly how to make it work lol...

so im gonna ignore this commit because there is some confusing stuff going on, but will make this change in master:

```
#ifndef ORANGE_NO_MSIL
        case ARCHITECTURE_MSIL:
            msil_examine_icode(head);
            break;
#endif
```

I looked at your other changes again and the 'compile to file' change in occparse.cpp wouldn't have worked so well for me since the optimizer and backend expect everything specified on the command line in one file, maybe I will look at making an ifdef change for it though since it seems like it could be a useful thing in some situations....

rochus-keller:
I still find it very strange why the windows build still fails even if I effectively reverted to [45bb903](https://github.com/LADSoft/OrangeC/pull/1032/commits/45bb9032834cac2f7e1ceefff45d71a0ddccc817), which was the last commit the windows build worked. 

Interestingly the windows build failed after 30 minutes (compared to the 5 minutes before my last commit). Maybe I'll be able to make sense of the huge amount of output from the build log and find a clue to the answer.

I would find it very desirable if my version of the frontend does not evolve away from yours, since it essentially does the same thing (except that I only want to compile a subset) and I dock my special code to the icf file. I could even live with all input files going into the same icf.

rochus-keller:
I close this and start over in a separate branch, with the new define up-front and minimal changes.

GitMensch:
Nearly exactly 30 minutes may be a build timeout from Appveyor, you'd see that in the end of the build log.

rochus-keller:
Unfortunately I don't know Appveyor, but I have now tried with a new, minimal pull request with only a few new ifndefs and effectively no change to the windows build and it stops after 30 minutes, just stating "Error: The operation was canceled."

PR #1041, opened 2024-07-15
Orange front and opt successfully build with GCC 4.8 (-std=c++11) on Linux

rochus-keller:
Having successfully migrated a C compiler to the ECS backend in the meantime (see https://github.com/rochus-keller/EiGen/tree/master/ecc), I now dare to (hopefully) do the same with OrangeC.

To make my work easier, I have now also migrated orangefront and orangeopt to C++11 so that I can compile them with BUSY and GCC 4.8 on my developer machine. I have successfully tested orangefront with a subset of the ECC testcases and assume that the migration did not introduce any bugs. 

Actually, it only needed a few changes to make it compatible with C++11, and in none of the changed parts did I have the impression that C++14 simplified the code (on the contrary, the code became even shorter with C++11).

I therefore recommend incorporating these changes into your code.

### DIFF
--- a/BUSY
+++ b/BUSY
@@ -68,6 +68,9 @@ let common : SourceSet {
         ./src/util/Utils.h
         ./src/util/xml.h
 
+        ./src/ocpp/Floating.cpp
+        ./src/ocpp/Floating.h
+
         ./src/occopt/config.cpp
         ./src/occopt/config.h
         ./src/occopt/configx86.cpp
@@ -143,7 +146,6 @@ let orangefront : Executable {
         ./src/occparse/unmangle.cpp
         ./src/occparse/wseh.cpp
         ./src/ocpp/Errors.cpp
-        ./src/ocpp/Floating.cpp
         ./src/ocpp/InputFile.cpp
         ./src/ocpp/MakeStubs.cpp
         ./src/ocpp/PipeArbitrator.cpp
@@ -213,7 +215,6 @@ let orangefront : Executable {
         ./src/occparse/winmode.h
         ./src/occparse/wseh.h
         ./src/ocpp/Errors.h
-        ./src/ocpp/Floating.h
         ./src/ocpp/InputFile.h
         ./src/ocpp/MakeStubs.h
         ./src/ocpp/PipeArbitrator.h
@@ -282,16 +283,11 @@ let orangeopt : Executable {
         ./src/occopt/istren.h
         ./src/occopt/localprotect.cpp
         ./src/occopt/localprotect.h
-        ./src/occopt/msilprocess.cpp
         ./src/occopt/optmain.cpp
         ./src/occopt/optmain.h
         ./src/occopt/optmodulerun.cpp
-        ./src/occopt/rewritemsil.cpp
-        ./src/occopt/rewritemsil.h
         ./src/occopt/rewritex86.cpp
         ./src/occopt/rewritex86.h
-
-        ./src/ocpp/Floating.cpp
      ]
 }
 
--- a/BUSY
+++ b/BUSY
@@ -7,8 +7,8 @@
 
 let config : Config {
     if (target_toolchain == `gcc) {
-        # NOTE: this build has been successfully run on Linux x86 with GCC 12 and Linux x64 with GCC 10
-        .lib_names += [ "m" "stdc++" ]
+        # NOTE: this build has been successfully run on Linux x86 with GCC 4.8 and 12, and Linux x64 with GCC 10
+        .lib_names += [ "stdc++" "m" ] # NOTE: order is relevant to avoid missing floor reference from stl
         if build_mode == `debug {
             .ldflags += "-shared-libgcc"
         } else {
--- a/BUSY
+++ b/BUSY
@@ -14,14 +14,14 @@ let config : Config {
         } else {
             .ldflags += "-static"
         }
-        .cflags_cc += "-std=c++14"
+        .cflags_cc += "-std=c++11"
     	.defines += "HAVE_UNISTD_H"
     }else if (target_toolchain == `clang) {
     	# NOTE: this is an experimental compile with Clang 4.0.1 and the libc++ headers;
     	# it compiles when the original C headers from the system are available
     	# linking is yet another adventure to be solved
         .lib_names += [ "m" "stdc++" ]
-        .cflags_cc += "-std=c++14";
+        .cflags_cc += "-std=c++11";
     	.defines += "HAVE_UNISTD_H"
         .include_dirs += [
             //home/me/Programme/clang-4/includes
--- a/src/occopt/optmain.cpp
+++ b/src/occopt/optmain.cpp
@@ -95,10 +95,12 @@ void* operator new(size_t aa)
     {
         // If malloc fails and there is a new_handler,
         // call it to try free up memory.
+#if __GNUC__ > 4
         std::new_handler nh = std::get_new_handler();
         if (nh)
             nh();
         else
+#endif
             throw std::bad_alloc();
     }
     rv->size = bb;
--- a/src/occparse/c.h
+++ b/src/occparse/c.h
@@ -82,7 +82,7 @@ public:
     SymbolTable<T>* Next() const { return next_; }
     void Chain(SymbolTable<T>* chain) { chain_ = chain; }
     SymbolTable<T>* Chain() const { return chain_; }
-    auto ReleaseNext() { auto rv = next_; if (next_) next_ = next_->next_; return rv; }
+    SymbolTable<T>* ReleaseNext() { SymbolTable<T>* rv = next_; if (next_) next_ = next_->next_; return rv; }
     int Block() const { return blockLevel_; }
     void Block(int level) { blockLevel_ = level; }
 
@@ -1251,9 +1251,7 @@ typedef struct _atomicData
 
 CONSTEXPR inline TYPE* basetype(TYPE* a)
 {
-    if (a)
-        a = a->rootType;
-    return a;
+    return a ? a->rootType : a;
 }
 
 CONSTEXPR inline bool __isref(TYPE* x) { return (x)->type == BasicType::lref_ || (x)->type == BasicType::rref_; }
--- a/src/occparse/ccerr.cpp
+++ b/src/occparse/ccerr.cpp
@@ -521,7 +521,7 @@ int printerr(int err, const char* file, int line, ...)
 
 bool RequiresDialect::Base(Dialect cpp, int err, const char* feature)
 {
-    static std::unordered_map<Dialect, const char*> lookup = {
+    static std::unordered_map<Dialect, const char*,EnumClassHash> lookup = {
         {Dialect::c89, "C89"},     {Dialect::c99, "C99"},     {Dialect::c11, "C11"},    {Dialect::c2x, "C23"},
         {Dialect::cpp11, "C++11"}, {Dialect::cpp14, "C++14"}, {Dialect::cpp17, "C++17"}};
     if (cpp >= Dialect::c89 && cpp < Dialect::cpp11)
--- a/src/occparse/compiler.h
+++ b/src/occparse/compiler.h
@@ -51,6 +51,15 @@
 #include "c.h"
 #include "beinterfdefs.h"
 
+struct EnumClassHash
+{
+    template <typename T>
+    std::size_t operator()(T t) const
+    {
+        return static_cast<std::size_t>(t);
+    }
+};
+
 #define M_LN2 0.693147180559945309417
 #define M_LN10 2.30258509299404568402
 
--- a/src/occparse/declcpp.cpp
+++ b/src/occparse/declcpp.cpp
@@ -3996,7 +3996,6 @@ bool ParseAttributeSpecifiers(LEXLIST** lex, SYMBOL* funcsp, bool always)
                         }
                         else
                         {
-                            using namespace std::literals;
                             std::string stripped_ver = StripUnderscores((std::string)(*lex)->data->value.s.a);
                             if (stripped_ver == occNamespace)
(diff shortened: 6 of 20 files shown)


LADSoft:
ok cool, i was wondering if you were still working on this?    I very much like how you took out all those std::make_unique calls, i never liked that but didn't know what the alternative was lol...   I'll test this and then if it works ok I'll incorporate it :smile:

@chuggafan there are a could of minor changes in ppPragma.* as well....

rochus-keller:
>  i was wondering if you were still working on this

I said that it will take some time, since I work on a lot of projects in parallel, and - in the present case - I realized that I still had to fill some gaps in my knowledge to have any chance of understanding your IR and its internal data structures and properly map them to Eigen IR. I hope that my knowledge is now sufficient - we'll see.

Currently I do experiments with the optimizer; the output IR looks pretty different, and the register allocator assumes a specific CPU architecture; so I currently evaluate on which IR (before or after optimizer) I should focus. I assume you did performance measurements. How big is the effect of the optimizer on performance?

LADSoft:
theoretically the register allocator can be taylored to an architecture based on a header file....   haven't done it yet so i don't know how well that works in practice.     For it to work you have to have a normal register file of some sort.   It really does two register files one for FP and one for integers.   It is also able to handle the difference between scratch registers and registers that are able to remain persistent across a function....

i haven't measured each of the optimizations individually, but, i do know debug code runs a lot slower (40-50%) and a big part of  that would likely be the fact that the register allocator is turned off,

rochus-keller:
> i do know debug code runs a lot slower (40-50%) and a big part of that would likely be the fact that the register allocator is turned off,

I doubt the register allocator has such a big influence. I see e.g. a factor two in performance between GCC without optimization and with enabled -O2 (see
https://github.com/rochus-keller/Oberon/blob/master/testcases/Are-we-fast-yet/Are-we-fast-yet_results.ods). The same factor I see with other C compilers without optimization (e.g. TCC). That means all the complex optimization machinery (of which the register allocator is only a small part) causes a factor two speed-up. 

I will see whether I manage to do some performance measurements with the official OrangeC installation on Windows using either the C++ version of the Awfy suite (see https://github.com/rochus-keller/Are-we-fast-yet/tree/main/Cpp) or the C99 code generated from the Oberon+ version (see http://software.rochus-keller.ch/Are-we-fast-yet_ObxIDE_Cgen_2021-12-30.zip). I think it's important to know what the optimizer is capable of doing.

LADSoft:
thank you!

rochus-keller:
I tried to build the mentioned benchmarks with OrangeC on my Windows11 x64 machine; with MSVC2015 and MinGW it worked (though I had to split the long string in Json.cpp in two for MSVC2015), but with OrangeC I get the build errors as listed in the two log files:

- http://software.rochus-keller.ch/OrangeC%20Windows%20awfy%20C99%20build%20errors.txt
- http://software.rochus-keller.ch/OrangeC%20Windows%20awfy%20c++98%20build%20error.txt

The OrangeC version from the official download still has the issue you fixed in Mai, so I used the version from here: https://github.com/LADSoft/OrangeC/releases/download/Nightly/ZippedBinaries.zip

What am I doing wrong?

LADSoft:
ok you aren't doing anything wrong.    

The c99 problem is well, apparently at this point in the evolution of the C/C++ language there isn't such a thing as a duplicate public, all publics are supposed to be in common data.    I totally missed that detail hence it doesn't work.

That will take me a couple of days to fix and vet...

The C++ problem is also on my end, apparently the error check doesn't handle constructors with member initializers.   I'll go ahead and fix it, not sure if it will be available tonite or not.

rochus-keller:
Thanks for the info. I'm glad you can fix it, and it doesn't hurry.

chuggafan:
So this is actually going against my current PR in the form of the
fact that I'm moving towards C++17 with the introduction of items such
as std::variant, nested namespaces, etc., and hopefully later we can
move towards things such as std::filesystem, my personal opinion is
that GCC 4.8.5 is, frankly, absurdly old at this point in the compiler
world and not particularly worth our time supporting as a direct
target for compilation. In terms of CPU platform I'd say supporting
old targets is different, but even GCC is moving forward past this.
When the Linux Kernel itself requires a minimum version of GCC of 5.1
because it supports C11[1] I think there's real concern that binding
ourselves that far back would be an issue.
I don't think it's worth supporting too old compilers and I think that
limiting ourselves here will make long-term development much more
difficult than necessary.
I will be testing my changes against the pulled down code in the
coming week starting today however, hopefully I can get it back to a
good state.

As for the register allocator being part of performance changes, I
would agree believe that the allocator is a huge amount of the
efficiency gains, I did compiler optimization switch selection at a
previous job of mine by programmatically changing all (12ish) compiler
options, some of the biggest optimizations were the absolute most
basic things, common subexpression elimination, the register
coalescer, and peephole optimization were among the most noticeable
impacts in terms of optimization. At the bottom of the pile were
things such as SIMD and software pipelining. Now, I didn't test each
switch together, but I did test two batches of "what I expected to be
the most impactful batch" where each turned on and off, and "the less
expected impact", and the two groups had performance in line with what
I expected. I would definitely believe that a better register
allocator, allowing less spills so that we don't have to reach for
stack memory as often, would be a significant speedup in any compiler.
This is why a *lot* of research still goes into register allocators as
best as I can tell from whenever I actually look up the topic.
[1] https://www.kernel.org/doc/html/latest/process/changes.html

rochus-keller:
> my personal opinion is that GCC 4.8.5 is, frankly, absurdly old at this point in the compiler world and not particularly worth our time supporting as a direct target for compilation

Why would you implement just another Clang? Why don't you contribute to Clang if you prefer modern modern C++ (or GCC in recent time)? I think one of the major core benefit propositions of OrangeC is exactly that it doesn't require a modern modern toolchain and doesn't produce executables of > 100 MB in size, and is therefore very attractive for embedded, low-resource and also legacy systems. In case of the present code base, which is huge and implemented with very moderate language feature requirements, it makes little sense, to migrate parts of it in hindsight to a more modern language version, thus giving up a most essential advantage of the product.

rochus-keller:
> I would agree believe that the allocator is a huge amount of the efficiency gains

Let's do measurements instead of speculating. As soon as OrangeC is able to compile the benchmarks we can measure the effect of the optimizer (or parts thereof).

rochus-keller:
> When the Linux Kernel itself requires a minimum version of GCC of 5.1 because it supports C11[1] 

Linux started to allow C11 in 2021, i.e. after 30 years of gcc89. The C++11 standard is from the same year as the C11 standard, isn't it? Obviously both, C11 and C++11 are good enough to build huge compilers and operating systems.

LADSoft:
I'm kinda on the fence about the C++11/later issue...    i see that it makes sense in one sense to keep it at the older standard but there are somethings it would be nice to use.    And honestly, for all the years I've been working on this there doesn't seem to be enough of any kind of code base to go catering to things, and as we well know compilers are a dime a dozen any more anyway, so, i see it as mostly it is for me and anyone who wants to work on it more than it is for a user base at this point.  So I'm not too inclined to limit development to only C++11 although if the code can be equally simple by preferring c++11 features that seems prefererrable.

LADSoft:
as far as getting the benchmarks to compile, the CD.CPP file stressed out the compiler in various ways....    which is good lol....    I've got the C++ version of the benchmarks compiling now but they crash.

 The C99 variant is probably going to be easier to fix than I originally thought so I'll try to wrap both up over the weekend....

rochus-keller:
Thanks for the update.

Regarding the C++ version: It is your project and of course you decide how to implement it. I just wanted to point out one of the main strengths of the project compared to similar projects, and to warn against following the sirens' songs too quickly and without necessity. Having a C++11 (or even C++14) available for older or low-resource systems is an immeasurable advantage. More recent C++ versions indulge with more syntactic sugar, but also significantly increase compiler dependency and complexity.

GitMensch:
Is there a chance to get at last the compile part into the OrangeC testsuite for CI?
Running the benchmark program and output its runtime would be nice as well, but I don't know how near this gets us to the timeout

LADSoft:
@rochus-keller, yes I see the advantage but, I'm just not getting feedback that people actually want a compiler that compiles on very old systems, other than you.    The only person who came close wasn't interested in compiling occ on older platforms, just running it there.

LADSoft:
@GitMensch 

as far as the timeout I'm hoping it is a little better whenever I can finally merge my c++17 working branch back in...  I saw very good numbers for speeding up the compiler compiling itself (23%) although that went down a bit again (to about 15%) when i switched back to compiling in C++17 mode.   Apparently, enabling c++17 in libcxx causes more stuff to be compiled...

that was running the build script in its entirety, just compiling only the compiler with itself is maybe 30% for the c++14 compile...

the main holdup firht now is im also adding filesystem header and that challenged the constexpr processing...   im still trying to get out from under it...

anyway yeah i was wondering if maybe i should compile/run the benchmark as well.    Luckily, it doesn't take very long to compile it....

rochus-keller:
As I said, it's your project and your decision. Don't see me as a stakeholder (so far I'm just an onlooker). But looking at your implementation I meant to recognize friendlyness to legacy systems as your main intention. But apparently I'm wrong.

> I'm just not getting feedback that people actually want a compiler that compiles on very old systems
Well, what's the need and core benefit for yet another C/C++ compiler at all? And why do you support DOS if not for that purpose?

LADSoft:
well just to set the record straight we aren't supporting dos any more either...   I don't need to see a specific need either for me to want to work on this :smile:

@chuggafan just so we are clear it is ok to continue planning to use C++17 features while developing orange C.

LADSoft:
@rochus-keller I put a version that compiles the C++ version of are-we-fast-yet in the repository.

At one point I had a version that compiled the C99 version as well but that was a disaster so i delayed finishing it.   Will get it fixed up before too long...

rochus-keller:
Ok, thank you very much for making it work. 

I run the benchmarks on a more recent laptop of mine under Windows 11 using MSVC2015, MinGW32 and OrangeC.

Here are the detailed results: [awfy++ windows x86 results.pdf](https://github.com/user-attachments/files/16354814/awfy%2B%2B.windows.x86.results.pdf)

Between the optimized versions built with OrangeC and MSVC2015 is a factor 1.5, and between OrangeC and MinGW32 a factor 1.9. 

Between the optimized and unoptimized version built with OrangeC is only a factor 1.14 (3.9 with MSVC2015 and 3.6 with MinGW32). The unoptimized version built with OrangeC is therefore much faster than the unoptimized versions built with MSVC and MinGW, but the OrangeC optimizer apparently has a small effect. 

So I could easily afford to directly use the output of orangefront and ignore orangeopt. I'm currently implementing a backend for cparser (without libfirm), where the unoptimized result is faster than that of GCC, and the optimizer also has a pretty little effect (see https://github.com/libfirm/libfirm/issues/37).

I'm a bit concerned that both, the C++ and the C version of awfy didn't immediately work with OrangeC and quite a lot of changes were necessary from your side to make it work.

rochus-keller:
This was in no way intended as a reproach or complaint. I merely wanted to share my thoughts and findings. I realize that this is an immense project.

LADSoft:
thank you.

I finally got the C99 version of the test suite to run (mostly); in addition to the BSS issues there were a couple of issues with the C99 structiore-initialization-as-expression that I hadn't found yet.   Once I got those fixed it was mostly smooth sailing....   main issue is the HAVLAK tests have a protection fault.   When I looked into it it seemed like those tests really use memory intensely and it was running out...   after noting that cl.exe has the same problem i opted not to deal with this right now and just commented out that test...

FWIW CLANG 16.0.6 has the same problem with not being able to duplicate public definitions in multiple modules as orange c did.

rochus-keller:
Great, thank you very much for fixing it. I will run the benchmarks asap. 

Concerning running out of memory: the generated C99 application requires the Boehm GC. There is a build.txt file included (at least I hope so) with the required define. Otherwise it's no surprise that there are memory issues.

Concerning Clang: I avoid this compiler as far as possible; but I was able to compile and run the C99 awfy with both chibicc and cparser (the former only with the updated generator), and gcc anyway. Because of the two, I probably got the wrong idea that the code is unproblematic. The C++ version is from my point of view quite conservative C++98 code; the local structures were the way closures were implemented back then, before C++11 introduced lambda expressions (which are in fact just syntactic sugar for the manual method as in my code).

rochus-keller:
Meanwhile I managed to run the C99 version of the benchmark. To use the Boehm GC, I had to create an orangec-compatible library file from the Boehm dll using oimplib.exe. That worked seamlessly. With MinGW I had to add the -fcommon option, because apparently since GCC 10 -fno-common is the default which leads to many duplicate symbol errors.

I run the benchmarks on the same laptop as the previous C++ runs, again using MSVC2015, MinGW32 and OrangeC.

Here are the detailed results: [awfy c99 x86 results.pdf](https://github.com/user-attachments/files/16602778/awfy.c99.x86.results.pdf)

Since the laptop is much faster than my development machine, the averages became too small; I instead compared the sums. 

For comparison I also added a sums based evaluation of the C++ runs: [awfy c++ windows x86 results.pdf](https://github.com/user-attachments/files/16602795/awfy.c%2B%2B.windows.x86.results.pdf) (see second page). Because the NBody sum value is available, I added it to the evaluation, which changed the overall results quite a bit (MSVC/Orange -O2 now 1.84 instead of 1.50).

Between the optimized versions built with OrangeC and MSVC2015 is a factor 1.63 (C99) and 1.84 (C++), and between OrangeC and MinGW32 a factor 1.57 (C99) and 2.13 (C++). C++ code compiled with OrangeC is therefore ~20% slower than C99 code.

Between the optimized and unoptimized version built with OrangeC is a factor 1.17 (C99) and 1.14 (C++).

PR #1069, opened 2025-04-12
Fix clang-cl std=c++17 command

chuggafan:
I forgot this in #810, I'm still confused because I'm seeing the build fails *occasionally* on NORMALcompiles.bat, and sometimes not?!?!? Also the first clangcompiles.bat I did on my machine worked fine!?!?!

*SIGH*, I really am in desperate need to get my build-box up.... Delays because it turns out HDDs are slow as molasses compared to SSDs and it really does affect performance for the build-box.


LADSoft:
so i ran it for a while last night then all day today...   what I'm expecting is that omake is having some problems with future that causes a throw.   Which leads me to think that we should put a try catch around all the main routines to catch things thrown by the runtime...    But there are bugs in that kind of try-catch so I'm off to write an issue so I can fix them....

LADSoft:
I made issue #1072 to deal with a build problem.    I don't know if it is related to this or not...

PR #1074, opened 2025-04-18
Fix path separators & 'copy' commands in makefiles, on Linux

tkchia:


chuggafan:
Hmmm, I like this change, 99% of the way done with linux support there. Thanks for the PR!

Only additional thing I have looked at previously to do with doing `omake fullbuild` beforehand is updating `copydir` since that is pure win32, and I'm not exactly sure how I would need to update it... I think for linux we might be able to do everything with `cp -r` for recursive copying? I'm going to be working on the `omake` updates again this weekend, see what I need to do about the switch commands, I'll probably do some tests then.

chuggafan:
I was actually working on this once I found your PR, I've found that `cp -r` is a good replacement except for one thing:
`cp` doesn't create directories, so `copydir` can just be made into a `mkdir` + a `cp` call.
I'd have to do a larger re-write in order to get the `clibs` side working because all of that is untranslated from `\` to `$(PATH2EXT)` and will likely be more complex even if I did nothing...

PR #1119, opened 2026-02-06
Fix omake to be able to be used everywhere

### LINKED ISSUE #1117: omake not using right path for self
I have the ORANGEC and ORANGEC_HOME environment variables set.   Sometimes when I run omake it uses the wrong path for itself for nested calls.

as an example (using cmd.exe as the shell)
```
c:\OrangeC\tests>omake clean
        c:\OrangeC\tests\omake /Calexcs clean
'c:\OrangeC\tests\omake' is not recognized as an internal or external command,
operable program or batch file.
```

if i manually add a path to omake, e.g.

```
\orangec\bin\omake
```

then things work...

chuggafan:
Uses the correct OS-specific processes to ensure OMake behaves properly no matter what directory we spawn omake from, or how we spawn omake

Fixes #1117

### DIFF
--- a/src/omake/os.cpp
+++ b/src/omake/os.cpp
@@ -438,17 +438,39 @@ std::string OS::SelfPath()
     char path[MAX_PATH];
     DWORD chars_written = GetModuleFileNameA(NULL, path, MAX_PATH);
     DWORD last_error = GetLastError();
-    if (chars_written && last_error != ERROR_INSUFFICIENT_BUFFER))
+    if (chars_written && last_error != ERROR_INSUFFICIENT_BUFFER)
     {
         return std::string(path);
     }
-    throw std::system_error(last_error);
+    throw std::system_error(last_error, std::system_category());
 #else
     char path[PATH_MAX];
-    ssize_t num_chars = readlink("/proc/self/exe", path, PATH_MAX);
+    char* readpath;
+    if (!access("/proc/self/exe", R_OK))
+    {
+        readpath = "/proc/self/exe";
+    }
+    else if (!access("/proc/curproc/file", R_OK))
+    {
+        readpath = "/proc/curproc/file";
+    }
+    else if (!access("/proc/self/path/a.out", R_OK))
+    {
+        readpath = "/proc/self/path/a.out";
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


src/omake/os.cpp
```diff
@@ -432,6 +432,25 @@ std::string OS::AbsPath(const std::string& str)
...
 }
+std::string OS::SelfPath()
+{
+#ifdef _WIN32
+    char path[MAX_PATH];
+    DWORD chars_written = GetModuleFileNameA(NULL, path, MAX_PATH);
+    if (GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+    {
+        return std::string(path);
+    }
```
GitMensch: Doesn't that miss a return/throw?

src/omake/os.cpp
```diff
@@ -432,6 +432,25 @@ std::string OS::AbsPath(const std::string& str)
...
+#else
+    char path[PATH_MAX];
+    ssize_t num_chars = readlink("/proc/self/exe", path, PATH_MAX);
+    if (num_chars == PATH_MAX)
```
GitMensch: possibly check -1 here

src/omake/os.cpp
GitMensch: Also: msdn says it may return another error (in that case chars_written is zero, so maybe `chars_written && GetModuleFileNameA !=...`)

GitMensch (changes requested):
Kind of nitpicking - but only because I like the change overall!

src/omake/os.cpp
chuggafan: No, the check is for num_chars because of POSIX readlink not putting a null terminator when you're longer than the possible path.
readlink is a *really* bad API that should be replaced with something safer.

chuggafan: I've roughly copied your suggestion on the different paths we can read from there, I made it so that if the end of PATH_MAX is not the null terminator then we throw (that way we can distinguish between the two possible cases there) and added the "-1" case.
This should be (mostly) sufficient, then?

LADSoft:
sorry im so far behind!   Thank you for fixing this :smile:

Issue #1137, opened 2026-03-25
OCC no longer accepts custom file extensions with `-x`


bradking:
OrangeC 7.0.7 fails to compile a custom source file extension:

```console
>occ -yy -x c main.zzz
occ (OrangeC) Version 7.0.7
...
main.zzz is not an object file
Fatal error: Fatal Error...Return code: 1
```

OrangeC 6.73.8 worked:

```console
>occ -yy -x c main.zzz
occ (OrangeC) Version 6.73.8
...
Return code: 0
>main
...
```

LADSoft:
this is actually the same as #1131.   I know why it broke, we went through and explicitly decided not to accept files without known extensions to match GCC, but i didn't think about the various switches that would force a C or C++ compile.    I will work on it soon.

LADSoft:
hi @chuggafan, if you want to look at this that would be great!    There are changes required to occparse though, it immediately sends any file that doesn't end in one of the C or C++ extensions on to the backend without compiling it (in osutil.cpp).

chuggafan:
Hmmm.... an easy fix (now that I know where I'm looking) is to insert a "found = true" if `prm_language.GetExists()` is true in `InsertOneFIle`, that would mean that every file in a row gets treated as whatever language we set it as, I think? gcc behavior is a little bit different, as seen in the following, do we want to support that?

```
~$ cat test.c
#include <stdio.h>
int thing();
int main()
{
        printf("Hello from test1 %d\n", thing());
}
~$ cat test2.c
#include <iostream>
int thing()
{
        std::cout << "Hello from test2" << std::endl;
        return 1;
}
~$ gcc -x c++ test2.c -x c test.c
/usr/bin/ld: /tmp/ccsSz0Ls.o: warning: relocation against `_ZSt4cout' in read-only section `.text'
/usr/bin/ld: /tmp/ccsSz0Ls.o: in function `thing()':
test2.c:(.text+0x11): undefined reference to `std::cout'
/usr/bin/ld: test2.c:(.text+0x19): undefined reference to `std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)'
/usr/bin/ld: test2.c:(.text+0x20): undefined reference to `std::basic_ostream<char, std::char_traits<char> >& std::endl<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&)'
[...]

Issue #1136, opened 2026-03-25
OCC no longer fails when linking a missing library


bradking:
OrangeC 7.0.7 silently succeeds when asked to link a missing library:

```console
>occ -yy main.c -lDoesNotExist
occ (OrangeC) Version 7.0.7
...
Return code: 0
```

OrangeC 6.73.8 failed as expected:

```console
>occ -yy main.c -lDoesNotExist
occ (OrangeC) Version 6.73.8
...
Error: Library 'DoesNotExist.l' does not exist or is not a library
        1 Errors, 0 Warnings
        Errors encountered, not creating output file
Return code: 1
```

LADSoft:
I'm not sure what happened here.   I'll lk into it...

LADSoft:
will fix this along with #1131/#1137

PR #1139, opened 2026-03-25
Fix #1138, check DBL ranges properly

### LINKED ISSUE #1138: OCC rejects `0` as initializer in array of `double` in C++
Given this source:

```console
>type data.cxx
int main(void)
{
  double data[] = { 0 };
  return 0;
}
```

OrangeC 7.0.7 fails to compile it:

```console
>occ -yy data.cxx
occ (OrangeC) Version 7.0.7
...
Error(366)    data.cxx(3):  Conversion from 'int' to 'double' would require narrowing
1 Errors
```

OrangeC 6.73.8 worked:

```console
>occ -yy data.cxx
occ (OrangeC) Version 6.73.8
...
Return code: 0
>data
...
```

chuggafan:


chuggafan:
I tried taking a look also at #1137, but didn't see anything pop out to me except that `occparse` behaves as expected, I wouldn't be surprised if #1136 is related to #1137.

LADSoft:
@chuggafan 

in #1137, like i said in the issue the parser isn't going to compile it at present without a proper extension...    I didn't get any further than that because it is easy to fix that issue but 1131 indicates a further detail:

#1136 is probably a separate issue, either the compiler looks it up and can't find it and won't pass it to the linker, or the linker discards it when it can't open it.    I would start by using occ -yyyy -Lundefinedlib.l  aa.c to see what is being passed out of the compiler...

aa.xx should compile to aa.xx.o not aa.o...   and there may be other output files where the .xx gets lost as well...

PR #1157, opened 2026-06-19
Fix issue #1156

### LINKED ISSUE #1156: OCC no longer fails on unknown flag
OrangeC 7.10.59 silently accepts an unknown compiler flag:

```console
>occ -c example.c -BadFlag && echo Success || echo Failure
occ (OrangeC) Version 7.10.59
...
Unknown switch '-B'
...
Success
```

OrangeC 7.0.7 rejected them:

```console
>occ -c example.c -BadFlag && echo Success || echo Failure
occ (OrangeC) Version 7.0.7
...
Unknown switch '-B'
...
Failure
```

chuggafan:
When a subcommand returns an error code, use an error code to indicate that we encountered an error.
Fixes #1156.

### DIFF
--- a/src/occ/occ.cpp
+++ b/src/occ/occ.cpp
@@ -532,7 +532,7 @@ int main(int argc, char* argv[]) MAINTRY
     }
     delete optimizerMem;
     if (rv == 255)  // means don't run the optimizer or backend
-        rv = 0;
+        rv = 1;
     return rv;
 }
 MAINCATCH
--- a/src/occopt/configx86.cpp
+++ b/src/occopt/configx86.cpp
@@ -109,7 +109,8 @@ Executable Mode:
     d - dll                             c = crtdll.dll
     g - windowing                       m = msvcrtdll.dll
     
-Optimization control: OPTIMIZATION_DESCRIPTION Flags: OPTMODULES_DESCRIPTION
+Optimization control:)help" "\n" OPTIMIZATION_DESCRIPTION "Flags:\n" OPTMODULES_DESCRIPTION
+    R"help(
     -fsyntax-only                  compile only, don't produce an output file
     -funsigned-char                'char' type is unsigned
     -std=xxxx                       specify the language standard to use


LADSoft:
@chuggafan, thank you for looking at this.   yeah that was my immediate thought that the raw strings were causing some kind of issue :)

FWIW I've kinda been playing with my GCSE changes that I made last spring.    Mostly it is working, just some aliasing problems...    code is a lot cleaner this way...

chuggafan:
Yhea, I'm currently investigating the correct pre-processor changes required to fix this issue, since this is an item that is technically mandated as a pre-processing token under 5.4 of the spec [lex.pptoken].

LADSoft:
from what you are saying it seems like maybe the processing for raw strings should be done in the preprocessor rather than the compiler; it could turn them into what the compiler sees as 'normal' strings (expanding escape sequences appropriately of course)?   But I don't know how that would play with things like compiling preprocessed files....

chuggafan:
According to the C++ spec, raw string literals are processed in the pre-processor and in the lexer. 
For the pre-processor, this is the quoted bit in n4713:

```
3 If the input stream has been parsed into preprocessing tokens up to a given character:
(3.1) — If the next character begins a sequence of characters that could be the prefix and initial double quote
of a raw string literal, such as R", the next preprocessing token shall be a raw string literal. Between
the initial and final double quote characters of the raw string, any transformations performed in phases
1 and 2 (universal-character-names and line splicing) are reverted; this reversion shall apply before any
d-char, r-char, or delimiting parenthesis is identified. The raw string literal is defined as the shortest
sequence of characters that matches the raw-string pattern
encoding-prefixopt R raw-string
```

It may actually be that this is intentional behavior as part of the C++ spec, confusingly, as the 5.2 section of "Phases of translation" has macro expansion as step 4.

I may need to sit and think on it in this case.... double check with the committee, if need be if this is intentional behavior. But it would make sense for it to not be the case as the raw string literal is considered to be a single preprocessing token thus isn't eligible for macro expansion.

chuggafan:
Hmmm.
I've come to the conclusion that I need something more advanced, because the encoding prefix is also technically part of the string "token" information as well.
`encoding-prefixopt R raw-string` is a thing, and is annoying for me to deal with as it's multi-line and the preprocessor isn't really designed to handle multi-line tokens.

LADSoft:
@chuggafan

so as far as the overall issue with tokenizing:   the 'preprocessor' module doesn't do any real tokenizing as strings are passed on to the compiler, and it tokenizes itself.    But, to solve this kind of issue one solution is to modify the routine that replaces macros to have a state saying not to replace macros, between the beginning of the raw string and the end?     Technically one would also want to do the same for trigraphs, but, I think trigraphs have been elided from later specs anyway so it might not be an issue if we didn't touch that one....

I just realized this issue is requested for merge, do you want me to do that now and we can deal with the real problem later?

chuggafan:
> @chuggafan
> 
> I just realized this issue is requested for merge, do you want me to do that now and we can deal with the real problem later?

Yhea, that works with me, ;).

But I do agree I can add a "don't macro expand state", the problem is that technically from the PP stance we should treat it as one whole "string token", at least that's how I was treating it, I was going to, on a normal string, treat individual string tokens normally, otherwise I'd kick the StringToken back and then add each line until the end delimiter to a total "stingBuilder" that would continuously build the StringToken together until the final complete StringToken output....
You might have a better plan than me for that.

LADSoft:
so if you want to turn it into a 'normal' string token one way is to reuse the fact that '\; is a line continuation so we have the capability for line continuation.    At the lowest level you would detect the raw string then turn on line continuation for the duration of the string.    But it would be a little different than 'normal' line continuation in that non-ascii characters have to be converted into escape sequences....   I also have minor concerns about getting line numbering right after the string but that should already be handled by the continuation mechanism....

LADSoft:
so i looked at the changes for this merge request...   I think (not sure) that your change for converting 255 to 1 is gonna upset the apple cart for things like preprocessing/generating make stubs.   If you do those without compiling it should probably have an exit code of 0, not 1....    also i suspect that some of the things that can be displayed by the parser (other than usage) probably ought have an exit code of 0 as well....    but I'm not really sure at this point..

chuggafan:
Hmmm... I should test on the `occ -M` side then, because we should differentiate things that have intentional effect (-M not doing anything should return 0) and failure (Unknown flag must return 1).
I'll retake a look later today then, make sure it's behavior we want.

chuggafan:
Sorry for the delayed check, you're right on the money on the issues.
Something I thought of in the meantime is that on unix the entire space above Return Code 128 is allowed for whatever, but that code has to be able to be translated to a `0` or `1` or something else before it hits the actual terminal, which is the issue.
I know on windows we can do some weird thing where we make `argv[0]` incorrect for, say, `occparse`, but I can't think of a good solution to this on Linux except via checking if we have an open pipe to `occ` from `occparse` then kicking it back up with a message....
We might be able to also just consume all of the arguments in `occ` and solve the problem there, and allow `occparse` etc. to be standalone.

LADSoft:
i was thinking, maybe the cleanest approach is just introduce a new variable somewhere that gets processed through the ilstream/ilunstream interface so that each of the programs can reference its value.    We set it to true in occparse to tell other programs they don't need to be processing stuff?   Then there is no confusion about what exactly OCCPARSE is returning, we can just have to return the correct value after that.

Another thing is just choose a random value like 250 as a flag( ok maybe it has to be 120 for linux), and add the real return value to it.    Anything >=250(120) would mean, occopt and occ and occil don't need to process the program.    But I don't think that is as clean as the other approach...

LADSoft:
Hi @Chuggafan - any thoughts on this?   I've had to make the patch to occmain locally since i can't presently get occ to compile itself...   but the issue with accepting this is doesn't really fix the root problem with returning the wrong error code.    Do you want to look at it further or should I work on that?

chuggafan:
I don't think this is ready for acceptance in that case, we can separate out my commit for fixing the help output separately.
I think it's possible this might be addressable by whatever approach GCC/Clang use, but trying to find out how they do that is esoteric at best.
The one location in which we actually return this -255 error code is `occparse`, perhaps from that situation we could make `occparse` the driver that calls `occopt` and `occ` and have it be that organizational structure and that alone would solve the issue?

LADSoft:
the reason it is structured the way it is, is that occ is diifferent for every architecture anyway, so we can just have different ones with different names lieing around.   If we make OCCPARSE the runner, there has to be a separate version of OCCPARSE for each architecture, which means recompiling it several times....

I still think if we just pass a variable through the ilsttream/ilunstream interface that would solve the problem nicely.

anyway yeah go ahead and remove the return value change from this pr then I'll accept it...   I guess we need to address both the main issue of this PR and the return value issue separately though...

Thanks...

